Issue #19, opened 2024-02-06
[Feature] Support WSI analysis


quentinblampey:
### High priority
- [x] Add a WSI reader (read a wsi into Xarray and then transform to a SpatialData object) 
- [x] Add whole tissue segmentation + conversion to a `shapely` polygon
- [x] Support multi-polygon ROI in segmentation (or anything that uses `ROI.KEY`)
- [x] Tile embedding in order to encode tile-wise phenotypical information (with different pretrained models e.g., ResNet or in-domain representation learning models). The could be used in a similar fashion to spatial transcriptomics but they will be much sparser.
- [x] Tile clustering in order to generate cluster-maps with similar phenotypical characteristics. Since sopa is slide specific, the clustering should also be slide specific. Any cohort-specific clustering should be made outside of Sopa
- [x] Add the new dependencies into a new `wsi` extra

### Low priority
- [x] Add `openslide` optionnal backend
- [x] Add `sjoin` in the API
- [x] Add some tutorials to the documentation

### Will not be added for now
- hover-net segmentation method (or [CellVit](https://github.com/TIO-IKIM/CellViT)?)
- Save `.zarr` without the image, but instead keep a pointer to the original slide

stergioc:
That could be a great feature add, thanks @quentinblampey for creating this issue! 

So here are some additional features that could also be considered:
- Tile embedding in order to encode tile-wise phenotypical information (with different pretrained models e.g., ResNet or in-domain representation learning models). The could be used in a similar fashion to spatial transcriptomics but they will be much sparser.
- Tile clustering in order to generate cluster-maps with similar phenotypical characteristics. This clustering might need to be cohort specific instead of slide specific (TBD).

I can take a look on how to implement the IO for WSI and come back with a suggestion. :)

quentinblampey:
What segmentation method should we use? CellViT? Hover-net?
@HakimBenkirane is also using something else, called [HistomicsTK](https://github.com/DigitalSlideArchive/HistomicsTK)

The method needs to have the following characteristics:
- few dependencies, easy to install (e.g., no openslide dependency)
- the method performs well, is well-validated and widely used
- not too slow

stergioc:
While trying to make the whole patch embedding pipeline faster I have added the openslide backend at PR #26 . After some tests it is clear that openslide is much faster as a backend. After a bit of digging I found this is issue that it might be related: https://github.com/Bayer-Group/tiffslide/issues/72, I am adding it here just for reference atm.

[...]

PR #21, opened 2024-02-08
WSI I/O and Tissue Segmentation with Otsu

stergioc:
Hey @quentinblampey and @loic-lb,

in this PR I am adding two functions towards the goals of #19 :

```python
sopa.io.histopathology.read_wsi(path: Path) -> SpatialData
``` 
This function is using tiffslide as backend to get a zarr store from propriatery wsi formats (support is subject tifffile/tiffslide support) and returns a `SpatialData` obj with a `MultiscaleSpatialImage` and the proper transformations configurations. OpenSlide could also be used however it is not supporting python3.10 atm.

```python 
sopa.segmentation.tissue.hsv_otsu(sdata: SpatialData, **seg_params)
```
This function is performing an hsv colorspace otsu based segmentation of the tissue using the lowest available level in the WSI and appends the tissue segmentation polygons in the `SpatialData` object.

I did not do extensive tests for the different filetypes and I did not update the dependencies (the tiffslide dependency needs to be resolved).

Let me know what you think. 🔥


quentinblampey:
Thanks @stergioc, it looks very nice

I have made some changes to the IO part:
- rename `read_wsi` into `wsi` for more consistency with the rest of the code. For instance, now the syntax is `sopa.io.wsi("CMU-1.ndpi")`
- the `tiffslide` library has to be imported inside the `wsi` function, else every it will be imported automatically for all users using `sopa.io`, even those who doesn't have the `wsi` dependencies, creating an error
- some minor naming changes to stay consistent with the rest of the code

We still need to add the dependencies into a new extra (I will do that in a future PR), and I need to check the tissue segmentation part

quentinblampey:
I just added `wsi_autoscale` which is similar but uses the automatic scales made by `spatialdata`. We will see in the future which one is more convenient: for now we can keep both and one day we will deprecate one

loic-lb:
Hello @stergioc ,

Thank you for the code. I do not have major comments, except a few improvements for the `read_wsi` (now `wsi` function) :

```python
def read_wsi(path: Path, patch_size: int | tuple[int, int] = (256, 256)) -> SpatialData:
    image_name = Path(path).absolute().name.split(".")[0]
    tiff = tiffslide.open_slide(path)
    img = xarray.open_zarr(
        tiff.zarr_group.store,
        consolidated=False,
        mask_and_scale=False,
    )

    patch_size = (patch_size, patch_size) if isinstance(patch_size, int) else patch_size

    tiff_metadata = {
        "properties": tiff.properties,
        "dimensions": tiff.dimensions,
        "level_count": tiff.level_count,
        "level_dimensions": tiff.level_dimensions,
        "level_downsamples": tiff.level_downsamples,
    }

    images = {}
    for i, k in enumerate(list(img.keys())):
        ap = k if k != "0" else ""

        simg = SpatialImage(
            img[k].transpose("S", "Y" + ap, "X" + ap),
            dims=("c", "y", "x"),
            attrs={"metadata": tiff_metadata},
        ).chunk({"c": 3, "y": patch_size[1], "x": patch_size[0]})

        sf = tiff.level_downsamples[i]
        images[f"scale{k}"] = Image2DModel.parse(
            simg,
            transformations={"pixels": Scale([sf, sf], axes=("x", "y"))},
            c_coords=("r", "g", "b"),
        )

    mimg = MultiscaleSpatialImage.from_dict(images)
    mimg.attrs = {"metadata": tiff_metadata}

    return SpatialData(images={image_name: mimg})
```
I added two elements that I think are necessary:

- first, let the user choose a custom `patch_size` as it is often done in other libraries
- second, link the metadata to the slide and not only to the scale which makes more sense to me (and makes the metadata more easily accessible). Ideally, it would have been great to link them directly to the SpatialData object, but I do not think it is possible.

I will let @quentinblampey check if these changes are consistent with the rest of the code, otherwise you can just skip them =)

quentinblampey:
Ahah I was just changing the way to access `metadata`, and saw your comment afterwards

quentinblampey:
Thanks for the tissue segmentation, it works very nicely
I also made a few changes:
- Added docstrings (can you complete the argument description @stergioc ?)
- If there is only one image in the spatialdata object, there is no need to provide the name of the image as an argument
- cv2 is imported inside the function, to prevent errors for users without the cv2 dependency
- The resulting shapes are saved in `sdata[ROI.KEY]` for consistency with the rest of the code. For instance, segmentation will be run automatically only on the tissue segmentation now, without any additional changes
- Added logging and other minor changes

stergioc:
Great stuff!! Thank you both for the feedback and for streamlining the code! 
I will write complete the docstrings later today :)

stergioc:
Ok I added the docstrings for the hsv_otsu. Just an extra comment on the segmentation function. I know that @loic-lb has a similar function implemented somewhere for tissue segmentation that was a bit more tailored for H-DAB stained images. I don't know if it makes sense to try to incorporate something like that here. Either as a separate function or as an additional functionality in this one with an extra argument that specifies the staining protocol that was used.

quentinblampey:
LGTM!

PR #22, opened 2024-02-12
Test new CI

quentinblampey:


quentinblampey:
CI is now working nicely:
- running pre-commit (flake8, among others)
- build is deploying doc only when releasing a new version, else only tests
- using python 3.10

PR #26, opened 2024-02-13
Adding patch embeddings

stergioc:
Related to #19 

I added a patch embedding functionality for WSIs. Specifically, patches are retrieved with a specific `patch_width` from a specific `magnification` level. Then they are pushed as batches of `batch_size` into one of three embedding models. The resulting embeddings are stored as an additional image within the spatialdata object with index the class of the embedding model. 

in order to test it you can run the following:

```python
import spatialdata_plot
from sopa.io import wsi
from sopa.segmentation.tissue import hsv_otsu
from sopa.embedding.patches import embed_wsi_patches

slide = wsi('https://openslide.cs.cmu.edu/download/openslide-testdata/Aperio/CMU-1.svs')
hsv_otsu(slide, "CMU-1")
embed_wsi_patches(slide, "CMU-1", "Resnet50Features", 10, 224, 64, num_workers=8)

slide.pl.render_images("Resnet50Features", channel=[1, 2, 3]).pl.show()
```

the output of the plotting function using only the first 3 channels of ResNet features looks like this:


quentinblampey:
Hello @stergioc, thanks for the PR!
I tried using a little bit more the `Patches2D` class, which has some convenient functionnalities. I also added an assert to check that the model we ask for is existing

I have a few comments, see below:

sopa/embedding/patches.py
```diff
@@ -0,0 +1,133 @@
...
+    elif tiff_metadata["properties"].get("tiffslide.mpp-x"):
+        obj2mpp = {80: 0.125, 40: 0.25, 20: 0.5, 10: 1.0, 5: 2.0}
+        mppdiff = []
+        for objpow, mpp in obj2mpp.items():
+            mppdiff += [abs(mpp - float(tiff_metadata["properties"].get("tiffslide.mpp-x")))]
+        idx = np.argmin([abs(mpp - 0.44177416504682804) for objpow, mpp in obj2mpp.items()])
+        mpp_obj = list(obj2mpp.keys())[idx]
+        downsample = mpp_obj / target_magnification
+    else:
+        log.error("Error retrieving the mpp for {{image_key}}, skipping tile extration.")
```
quentinblampey, line 42: Here `image_key` is not accessible, so we can't log it

quentinblampey, line 42: Here `image_key` is not accessible, so we can't log it

sopa/embedding/patches.py
```diff
@@ -0,0 +1,133 @@
...
+):
+    image_key = get_key(sdata, "images", image_key)
+    image = sdata.images[image_key]
+    tiff_metadata = image.attrs["metadata"]
+
+    embedder = embed_batch(model_name=model_name, device=device)
+
+    level, resize_f, tile_s = _get_extraction_parameters(tiff_metadata, magnification, patch_width)
+
+    slide_dim = tiff_metadata["level_dimensions"][0]
```
quentinblampey, line 108: Unused variable. Do you want to do something with it or we delete it?

sopa/embedding/patches.py
```diff
@@ -0,0 +1,133 @@
...
+    return _
+
+
+def patch_embedding(
+    sdata: SpatialData,
+    image_key: str | None,
+    model_name: str,
+    magnification: float | int,
+    patch_width: float | int,
+    patch_overlap: float | int = 0,
```
quentinblampey, line 95: Unused argument, do we need patch overlap?

sopa/embedding/patches.py
```diff
@@ -0,0 +1,133 @@
...
+from spatialdata.models import Image2DModel
+from spatialdata.transformations import Scale
+
+from sopa._sdata import get_key
+from sopa.segmentation import Patches2D
+
+log = logging.getLogger(__name__)
+
+
+def _get_best_level_for_downsample(level_downsamples, downsample, epsilon=0):
```
quentinblampey, line 18: Can you add some typing? Same for the below function

sopa/embedding/patches.py
```diff
@@ -0,0 +1,133 @@
...
+        get_batches = [da.delayed(_get_patch)(image, box, level, resize_f) for box in patch_boxes]
+        batch = da.compute(*get_batches, num_workers=num_workers)
+        embedding = embedder(np.stack(batch))
+
+        xy = np.array([patches.pair_indices(k) for k in range(i, i + batch_size)]).T
+        output[:, xy[1], xy[0]] = embedding.T
+
+    embedding_image = SpatialImage(output, dims=("c", "y", "x"))
+    embedding_image = Image2DModel.parse(
+        embedding_image, transformations={"pixels": Scale([tile_s, tile_s], axes=("x", "y"))}
```
quentinblampey, line 126: Here "pixels" is hardcoded. Instead, we should use the intrinsic coordinate system of the image. I'll add it myself later

sopa/embedding/patches.py
```diff
@@ -0,0 +1,133 @@
...
+    tile_s = int(patch_width * downsample)
+
+    return level, resize_f, tile_s
+
+
+def _get_patch(image: SpatialImage, box: list, level: int, resize_f: float):
+    import cv2
+
+    patch = bounding_box_query(image, ("y", "x"), box[:2][::-1], box[2:][::-1], "pixels")
+    np_patch = np.array(patch[f"scale{level}"]["image"].transpose("y", "x", "c"))
```
quentinblampey, line 58: `"image"` is hardcoded, we have to use `next(iter(patch[f"scale{level}"]))` to get the first (and unique) DataArray of the Dataset

sopa/embedding/patches.py
```diff
@@ -0,0 +1,133 @@
...
+    """return the best level for a given downsampling factor"""
+    if downsample <= 1.0:
+        return 0
+    for level, ds in enumerate(level_downsamples):
+        if ds > downsample + epsilon:
+            return level - 1
+    return len(level_downsamples) - 1
+
+
+def _get_extraction_parameters(tiff_metadata, target_magnification, patch_width):
```
quentinblampey, line 28: Can you add a small docstring?

sopa/embedding/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .resnet import Resnet50Features
+from .histoSSL import histoSSLFeatures
```
quentinblampey, line 2: For consistency, I would rename the module `histo_ssl`, and the class `HistoSSLFeatures`, what do you think?

quentinblampey (changes requested):
I added some comments :)

sopa/embedding/patches.py
stergioc: What if I put this log after the call of the function and check if it returns false, instead of putting it here?

quentinblampey: Yes, sounds good! Indeed it will avoid passing the argument to this function

sopa/embedding/patches.py
```diff
@@ -0,0 +1,150 @@
...
+    it returns the best scale to get it from (level), a resize factor (resize_factor)
+    and the corresponding patch size at scale0 (patch_width)
+    """
+    if tiff_metadata["properties"].get("tiffslide.objective-power"):
+        downsample = (
+            int(tiff_metadata["properties"].get("tiffslide.objective-power")) / target_magnification
+        )
+    elif tiff_metadata["properties"].get("tiffslide.mpp-x"):
+        obj2mpp = {80: 0.125, 40: 0.25, 20: 0.5, 10: 1.0, 5: 2.0}
+        mppdiff = []
```
quentinblampey, line 44: What is the purpose of `mppdiff`? It seems to be also unused

stergioc: Hmm that is a good question. Something looks weird here. I will take a closer look.

quentinblampey: Ok thanks!

quentinblampey:
Just some minor changes needed but it already looks very good!

quentinblampey:
One minor question: what is the meaning of `obj`? "Objective"?

quentinblampey:
It's looking better, but we still have a slight shift in the plot. I'll investigate on this later
image

quentinblampey (approved):

quentinblampey:
Okay it's better now @stergioc. We still have a small shift (see below), but this is only a plotting issue. Indeed, by default matplotlib show a pixel centered (from -0.5 to 0.5), so there is a shift of 0.5 pixels. Usually, this is not visible, but here I have a transformation of `Scale(1000, 1000)`, so it makes a shift of 500 pixels during the plot. I opened an issue in spatialdata-plot here: https://github.com/scverse/spatialdata-plot/issues/216

image

sopa/embedding/patches.py
```diff
@@ -171,6 +172,8 @@ def embed_wsi_patches(
     embedding_image.coords["y"] = patch_width * embedding_image.coords["y"]
     embedding_image.coords["x"] = patch_width * embedding_image.coords["x"]
 
-    sdata.add_image(model_name, embedding_image)
+    shape_key = f"sopa_{model_name}"
+
+    sdata.add_image(shape_key, embedding_image)
 
-    log.info(f"Tissue segmentation saved in sdata['{model_name}']")
+    log.info(f"Tissue segmentation saved in sdata['{shape_key}']")
```
stergioc, line 179: Just a quick comment since I see you are changing things. This info message should write something like, WSI embeddings saved in ...

quentinblampey:
Of course the shift is less visible with a smaller `patch_width`. I just did some final clean up (docstrings, typing, ...). Ready for merging when I have your approval @stergioc 

image

stergioc:
Amazing @quentinblampey! I am running some tests on a bunch of in-house slides and testing the embedding speed. I will let you know when it is done :)

sopa/io/histopathology.py
```diff
@@ -99,7 +99,7 @@ def _default_image_models_kwargs(image_models_kwargs: dict | None) -> dict:
 def _open_wsi(path: str | Path) -> tuple[str, xarray.Dataset, Any, dict]:
     import tiffslide
 
-    image_name = Path(path).absolute().name.split(".")[0]
+    image_name = Path(path).stem
```
quentinblampey, line 102: Does it happen to have multiple extensions (like `.ome.tif`) or is it always one? Because `.stem` only removes the "last" extension, i.e. `image.ome.tif` would become `image.ome`

stergioc: hmm I did this because there are some filename that include `.` before the file extension. Here is an example from TCGA:  `TCGA-73-4658-01A-01-BS1.2859e392-66d8-407e-85db-25bf4e4c5ff3.svs` but there are more examples like this at the in-house databases with more segments in the filename that are separated with a `.`. Maybe we need a filetype detection function potentially the detect_format function of tiffslide can help?

quentinblampey: Maybe we could implement something like this, yes. Do you have a list of possible WSI extensions? Mostly, which one has a "double" extension?

But maybe we can keep your current implementation for now and open a new PR later if needed?

stergioc:
**Note:** at the moment the current implementation of the  OpenSlideStore works only for `zarr==2.14.2`. Not sure atm what would require in order to bump it to the current latest version of `zarr==2.17.0`.

stergioc:
I reverted the merge with wsi-openslide in order to first fix some of the issues that I identified before merging it with the wsi branch. We can stick with the tiffslide at the moment here.

quentinblampey:
Looking good to me @stergioc! Indeed we can do a PR later for openslide. I can merge now if it's all good on your side!

stergioc:
Go ahead yes it's all good from my side.

PR #51, opened 2024-04-09
Adds openslide backend

stergioc:
This is adding the openslide backend for reading WSIs. It is using a modified version of the OpenSlideStore class from  [manzt/napari-lazy-openslide/napari_lazy_openslide/store.py](https://github.com/manzt/napari-lazy-openslide/blob/main/napari_lazy_openslide/store.py) to do so. 

The issue related to stergioc/sopa#3 has been solved. At the moment, the feature extraction excecution time is x3 faster with the openslide backend (related to stergioc/sopa#2).

I let you @quentinblampey do the final checks :)


sopa/io/reader/wsi.py
```diff
@@ -12,19 +12,20 @@
 
 
 def wsi(
-    path: str | Path, chunks: tuple[int, int, int] = (3, 256, 256), as_image: bool = False
+    path: str | Path, chunks: tuple[int, int, int] = (3, 256, 256), as_image: bool = False, backend: str = "openslide"
```
quentinblampey, line 15: Since openslide is complex to install (contrary to tiffslide that is inside the 'wsi' extra), shouldn't we have tiffslide as a default?

sopa/io/reader/wsi.py
```diff
@@ -48,15 +49,14 @@ def wsi(
         images[f"scale{key}"] = scale_image
 
     multiscale_image = MultiscaleSpatialImage.from_dict(images)
-    multiscale_image.attrs["metadata"] = tiff_metadata
+    sdata = SpatialData(images={image_name: multiscale_image})
```
quentinblampey, line 52: When `as_image is True`, we will have the transformation already set inside the attrs of the image (see below). This might lead to some issues, but we can probably keep it like this for now

quentinblampey (commented):
Thanks for the PR, minor comments!

sopa/io/reader/wsi.py
stergioc: sure we can select the tiffslide as default

sopa/io/reader/wsi.py
```diff
@@ -35,10 +36,10 @@ def wsi(
             dims=("c", "y", "x"),
         ).chunk(chunks)
 
-        scale_factor = tiff.level_downsamples[level]
+        scale_factor = slide.level_downsamples[level]
 
         scale_image = Image2DModel.parse(
-            scale_image,
+            scale_image[:3, :, :],
```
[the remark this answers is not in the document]
stergioc: I noticed it when I used openslide backend. I guess the additional channel is the alpha channel in the RGBA colorspace. Not sure if it is library specific or file type specific. I could take a look on that

sopa/io/reader/wsi.py
stergioc: I am not sure I understand this comment here :|

quentinblampey: Don't worry it's a very minor thing

quentinblampey:
Looks good to me @stergioc, merging!

PR #90, opened 2024-07-12
Improvements for Cellpose segmentation 

pakiessling:
Made the following changes:

- Cellpose weights are now correctly used, previously Cyto3 was always applied instead
- Added an optional parameter clahe_kernel_size for skimage.exposure.equalize_adapthist which can be relevant for very big cells or patches. The default still is 1/8 of patch width and height
- Blur and CLAHE can be disabled by setting the parameter to 0
- If segmentation is run on multiple channels (e.g Nucleus marker + cell boundary marker) blur and CLAHE are now run on every channel separately which prevents spillover of intensities

### DIFF
--- a/sopa/cli/segmentation.py
+++ b/sopa/cli/segmentation.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import ast
+from typing import Iterable, Optional, Union
 
 import typer
 
@@ -25,6 +26,10 @@ def cellpose(
     ),
     clip_limit: float = typer.Option(
         0.2,
+        help="Parameter for skimage.exposure.equalize_adapthist (applied before running cellpose)",
+    ),
+    clahe_kernel_size: Optional[Union[int, Iterable[int]]] = typer.Option(
+        None,
         help="Parameter for skimage.exposure.equalize_adapthist (applied before running cellpose)",
     ),
     gaussian_sigma: float = typer.Option(
@@ -72,6 +77,7 @@ def cellpose(
         channels,
         min_area,
         clip_limit,
+        clahe_kernel_size,
         gaussian_sigma,
         patch_index,
         patch_dir,
@@ -96,6 +102,10 @@ def generic_staining(
     clip_limit: float = typer.Option(
         0.2,
         help="Parameter for skimage.exposure.equalize_adapthist (applied before running the segmentation method)",
+    ),
+    clahe_kernel_size: Optional[Union[int, Iterable[int]]] = typer.Option(
+        None,
+        help="Parameter for skimage.exposure.equalize_adapthist (applied before running cellpose)",
     ),
     gaussian_sigma: float = typer.Option(
         1,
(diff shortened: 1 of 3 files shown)


sopa/segmentation/stainings.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         image_key: str | None = None,
         min_area: float = 0,
         clip_limit: float = 0.2,
+        clahe_kernel_size: Optional[Union[int, Iterable[int]]] = None,
```
quentinblampey, line 34: Can you add docstrings for this parameter?

sopa/segmentation/stainings.py
```diff
@@ -105,8 +107,29 @@ def _run_patch(self, patch: Polygon) -> list[Polygon]:
             y=slice(bounds[1], bounds[3]),
         ).values
 
-        image = gaussian_filter(image, sigma=self.gaussian_sigma)
-        image = exposure.equalize_adapthist(image, clip_limit=self.clip_limit)
+        if len(self.channels) == 1:
```
quentinblampey, line 110: I think we don't need this "if-else" statement. The image always has a shape (C, Y, X), with sometimes C=1. For readability, I would keep only the "multi-channel" case below

sopa/segmentation/stainings.py
quentinblampey: And can you use the python>=3.10 typing syntax? I.e., `int | Iterable[int] | None`

sopa/cli/segmentation.py
```diff
@@ -27,6 +28,10 @@ def cellpose(
         0.2,
         help="Parameter for skimage.exposure.equalize_adapthist (applied before running cellpose)",
     ),
+    clahe_kernel_size: Optional[Union[int, Iterable[int]]] = typer.Option(
```
quentinblampey, line 31: Unfortunately, `typer` (the CLI tool) currently doesn't support Union types. We will not be able to pass an `Iterable[int]` to the CLI...Only the API will work with `int | Iterable[int] | None`

sopa/cli/segmentation.py
```diff
@@ -97,6 +103,10 @@ def generic_staining(
         0.2,
         help="Parameter for skimage.exposure.equalize_adapthist (applied before running the segmentation method)",
     ),
+    clahe_kernel_size: Optional[Union[int, Iterable[int]]] = typer.Option(
```
quentinblampey, line 106: You need to pass this parameter to `StainingSegmentation` line 175

quentinblampey (changes requested):
Thanks @pakiessling, looking very good overall, I added a few comments!

sopa/cli/segmentation.py
pakiessling: okay, what should the typing be? just int?

quentinblampey: Yes, just int (and `None` as a default value)

quentinblampey (approved):

quentinblampey:
All the tests passed, I'm merging, thanks again @pakiessling 🎉

PR #126, opened 2024-09-24
Fix compatibility with latest version of Baysor

lguerard:
Latest version of Baysor outputs either `segmentation_polygons2D.json` or `segmentation_polygons3D.json` depending on input data. This very basic fix would reestablish compatibility, in case users aren't running both 2D and 3D and outputting results in same folder

### DIFF
--- a/sopa/io/reader/macsima.py
+++ b/sopa/io/reader/macsima.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
 import logging
+import re
 from pathlib import Path
 
 from spatialdata import SpatialData
 
-from .utils import _general_tif_directory_reader
+from .utils import _deduplicate_names, _general_tif_directory_reader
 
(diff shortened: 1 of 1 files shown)


lguerard:
Actually, there's way more changes inside the Json output. The format completely changed, another option is needed to be passed to Baysor. I'll try it and come back here if I manage to have some fix(es).

quentinblampey:
Thanks for diving into this! Let me know if you need some help
It indeed seems that there are several issues with the new Baysor version...

quentinblampey:
Sorry @lguerard I don't know what happened, it automatically closed the PR when I wanted to push some stuff to your PR
For some reasons I couldn't push, even though being up to date with your changes

lguerard:
No worries at all, but it might not yet be fully working, it's quite WIP at the moment...

PR #127, opened 2024-09-25
Support Baysor>=7.0.0

quentinblampey:
Hi @lguerard, I added a test to check if the cell IDs are string
In the new 0.7.0 version, the cell IDs are strings, and we need to use the obs_names instead of CellID to compare to the loom matrix

Let me know if this works for you!

### DIFF
--- a/docs/tutorials/api_usage.ipynb
+++ b/docs/tutorials/api_usage.ipynb
@@ -78,8 +78,10 @@
    "outputs": [],
    "source": [
     "image_key = \"image\"\n",
-    "points_key = \"transcripts\" # (ignore this for multiplex imaging)\n",
-    "gene_column = \"genes\" # (optional) column of sdata[points_key] containing the gene names"
+    "points_key = \"transcripts\"  # (ignore this for multiplex imaging)\n",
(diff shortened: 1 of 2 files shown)


lguerard:
Thanks a lot for looking into that ! 

And thank you for including the `dask` backend so make it easier to do parallelization. I just tried with `write_patches_cells` but I'm getting an error, I'm not familiar at all with `dask` so I'm not sure what it means..

```log
RuntimeError: Can't redefine method: __getstate__ on class: __torch__.torch.utils.mkldnn.MkldnnBatchNorm (of Python compilation unit at: 000001A8C4E17ED0)
```

quentinblampey:
Could you open an issue with a detailed error log?
Since I'll merge this PR soon, it will be better to keep track of this issue

lguerard:
I'm reinstalling my environment right now, I'll create an issue if I encounter it again. 

But anyway, you removed the `settings` file and the dask support for now no ?

quentinblampey:
Merging this PR, thanks @lguerard!
And please let me know regarding the issue with the dask parallelization (see my comment above)

PR #142, opened 2024-10-22
Update config.yaml

wlason:
This is how I would suggest to update the slurm config.yaml

It adds a sensible default number of threads for HPC and fixes the small typo in specifying the `cluster` command.

It also creates logs which are prefixed with rule name (similarly to what snakemake would do if you run it with `--slurm` instead of specifying `--profile slurm` - if you do that, it would natively save to `slurm_logs/rule_{rule}/%j.log` in `.snakemake` directory), which are easier to read in case of errors - it allowed me to work out which rule needed more memory, for example.

### DIFF
--- a/workflow/slurm/config.yaml
+++ b/workflow/slurm/config.yaml
@@ -1,19 +1,22 @@
 cluster:
   "sbatch
     --partition={resources.partition}
-    --cpus-per-task={threads}
+    --cpus-per-task={resources.threads}
     --mem={resources.mem_mb}
     --job-name=smk-{rule}
     --gres=gpu:{resources.gpu}
-    --output=logs/%j" # you can update where the logs will be saved (absolute path, or relative to /.../sopa/workflow)
+    --time={resources.time}
+    --output=logs/{rule}_%j.log" # you can update where the logs will be saved (absolute path, or relative to /.../sopa/workflow)
 default-resources:
-  - partition=shortq
-  - mem_mb=32000
+  - partition=shortq # use the partition/queue name on your cluster
+  - mem_mb=15000
   - gpu=0
+  - threads=8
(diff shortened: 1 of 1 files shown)


quentinblampey:
Thanks @wlason, this looks good to be!
I'm currently out-of-office, I'll check that it also work on our clusters as an additional check, and then merge it 👍

workflow/slurm/config.yaml
```diff
@@ -1,19 +1,22 @@
...
+    --time={resources.time}
+    --output=logs/{rule}_%j.log" # you can update where the logs will be saved (absolute path, or relative to /.../sopa/workflow)
 default-resources:
-  - partition=shortq
-  - mem_mb=32000
+  - partition=shortq # use the partition/queue name on your cluster
+  - mem_mb=15000
   - gpu=0
+  - threads=8
+  - time="30:00:00"
```
quentinblampey, line 15: Minor comment: this might be something that doesn't fit well for all users. I think this is something that the user should add if needed. WDYT?

quentinblampey:
Hello @wlason, sorry for the delay, I had unexpected deadlines.
Can you update your PR as asked in the comment above? (basically, keeping the changes until line 11 included).
Then I'll merge it :)

quentinblampey:
Hi @wlason, I now released the version `sopa==2.0.0`! It changes slightly how the profiles work (see the new profiles [here](https://github.com/gustaveroussy/sopa/tree/master/workflow/profile)), so I'm closing this PR because it is now outdated unfortunately...
Still, don't hesitate to check the new version and its corresponding [migration guide](https://github.com/gustaveroussy/sopa/discussions/138), or the new doc, and let me know if you feel a similar PR would be needed :)

PR #159, opened 2024-11-29
xarray bump

ArneDefauw:
Sopa can not be installed with new xarray version xarray 2024.11.0 (due to from datatree import DataTree no longer working). Because of this it can also not be installed with new version of spatialdata 0.2.6. This is small PR to fix these issues.


quentinblampey:
Hi @ArneDefauw, I released `sopa==1.1.6`, this should be fixed now!

Issue #187, opened 2025-01-29
[Bug] `generic-segmentation` command seems to be broken


jeffquinn-msk:
## Description

The `generic-segmentation` command seems to be broken on v2.0.0

## Reproducing the issue

```
$ python -c "import sopa; print(sopa.__version__)"
2.0.0
$ sopa segmentation generic-staining --method-name cellpose --channels DAPI sdata.zarr
...
TypeError: stardist() missing 1 required positional argument: 'sdata'
$ sopa segmentation generic-staining --method-name cellpose --method-kwargs '{"diameter":6.5}' --c
hannels DAPI sdata.zarr
...
TypeError: cellpose() missing 1 required positional argument: 'sdata'
```

## Expected behavior

Seems like `sdata` needs to be passed into the dynamically resolved function.

I'm not sure what the intention here was:

https://github.com/gustaveroussy/sopa/blob/a972170a3d973b39995990c43372e9a95ba71c55/sopa/cli/segmentation.py#L165

but this seems like it will only work with the `*_patch` methods, which return some kind of curried function.

The `sopa segmentation cellpose` cli entrypoint passes the `cellpose_patch` method into the function so that's why that command works.

Trying to pass the patch method directly fails, because its not a default export in `segmentation/__init__.py` (except for cellpose_patch).

```
$ sopa segmentation generic-staining --method-name stardist_patch --channels DAPI sdata.zarr
AssertionError: 'stardist_patch' is not a valid method builder under `sopa.segmentation.methods`
```

## System

```
$ python --version
Python 3.10.13
```

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

```
$ pip freeze
aiobotocore==2.19.0
aiohappyeyeballs==2.4.4
aiohttp==3.11.11
aioitertools==0.12.0
aiosignal==1.3.2
anndata==0.11.3
[...]

PR #189, opened 2025-01-29
Enable Using Stardist from the CLI

jeffquinn-msk:
See https://github.com/gustaveroussy/sopa/issues/187


quentinblampey:
Looks good to me @jeffquinn-msk, thanks! Indeed, we should provide `stardist_patch` as a method name, which is not very intuitive. I'll update the doc to clarify this

quentinblampey:
I also made a few changes (see my commit mentioned above) to make it easier to run stardist via the CLI.
For instance, now you can run stardist like so:
`sopa segmentation generic-staining data.zarr --method-name stardist_patch`

PR #190, opened 2025-01-29
Improve Handling of Baysor Subprocess

jeffquinn-msk:
A few things I noticed with this

1) The way the `baysor` subprocess is being called, the exit code is not being propagated correctly. `sopa` will exit with status code 1 regardless of `baysor` exit code. When using a workflow manager like nextflow, the exit code can determine retry eligibility. Exit code 1 is usually never eligible for a retry because it implies a logic error but exit code 137 (OOM) for example would be eligible for a retry, so this diff will call `sys.exit` with the `baysor` subprocess exit code when we exit the python interpreter.

2) Baysor can run for a long time and I find myself wondering what's going on with it. To that end it would be nice to stream the stdout/stderr to the parent process stdout/stderr as its running. This diff adds a utility method `run_process_with_streaming_output` to do so and replaces subprocess.run with this utility method

### DIFF
--- a/sopa/cli/segmentation.py
+++ b/sopa/cli/segmentation.py
@@ -1,7 +1,7 @@
 import ast
 import sys
-from typing import Iterable
 from subprocess import CalledProcessError
+from typing import Iterable
 
 import typer
 
--- a/sopa/cli/segmentation.py
+++ b/sopa/cli/segmentation.py
@@ -1,5 +1,7 @@
 import ast
+import sys
 from typing import Iterable
+from subprocess import CalledProcessError
 
 import typer
 
@@ -245,7 +247,10 @@ def baysor(
 
     sdata = read_zarr_standardized(sdata_path)
 
-    baysor(sdata, config=config, min_area=min_area, patch_index=patch_index, scale=scale)
+    try:
+        baysor(sdata, config=config, min_area=min_area, patch_index=patch_index, scale=scale)
+    except CalledProcessError as e:
+        sys.exit(e.returncode)
 
     _log_whether_to_resolve(patch_index)
 
--- a/sopa/segmentation/methods/_baysor.py
+++ b/sopa/segmentation/methods/_baysor.py
@@ -2,8 +2,6 @@ import logging
 from functools import partial
 from pathlib import Path
 from subprocess import CalledProcessError
-from threading import Thread
-from queue import Queue
 
 from spatialdata import SpatialData
 
@@ -101,10 +99,7 @@ class BaysorPatch:
 
         _copy_segmentation_config(patch_dir / SopaFiles.TOML_CONFIG_FILE, self.config)
 
-        result = run_process_with_streaming_output(
-            self.baysor_command.split(),
-            cwd=patch_dir
-        )
+        result = run_process_with_streaming_output(self.baysor_command.split(), cwd=patch_dir)
 
         if result.returncode != 0:
             message = f"Baysor error on patch {patch_dir.resolve()} with command `{self.baysor_command}`"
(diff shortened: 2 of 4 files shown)


quentinblampey:
Hi @jeffquinn-msk, these are interesting comments. I have a few questions before going deeper into this:
- Using the right returncode is indeed very important. Just to understand what you did: is `CalledProcessError(returncode=result.returncode, ...)` enough to return the right code, or do you also need this `run_process_with_streaming_output` function? The latter function seems complicated, I need to take some time to better understand what it does.
- Regarding nextflow, do you really want to see the logs of all processes? We may have 30+ baysor processes in parallel, so the logs could become messy. NB: I'm not a nextflow user, so this is a naive question, maybe nextflow handles it nicely.
- Since you already use Nextflow, would you be interested in helping moving the Sopa pipeline to nextflow?

Thanks!

quentinblampey:
Hi @jeffquinn-msk, thanks for the extra details. I got some errors when using the API related to buffers, so I tried to dive into the code and I feel like there is a shorter / safer way to do the same. I opened a PR here: https://github.com/gustaveroussy/sopa/pull/199

In brief, it does the following.
- Baysor is logging outputs only if we run only one patch. This way, you'll see the logs in Nextflow (since one process runs one patch), but we still don't see them when running Baysor in parallel on all patches in the API.
- The return code is directly sent via `sys.exit` after running `subprocess` instead of catching the error in the CLI only
Let me know if something is missing from your feature request!

And, yes, once your nextflow pipeline is ready, I'll be happy to have a look! Meanwhile, some people from my lab may also help making a general nextflow pipeline for Sopa

quentinblampey:
Yes, here is the log. It doesn't actually stop the process: it successfully runs segmentation, it's only printed to the console, but it may be a sign of other issues.

```sh
Exception in thread Thread-25 (connect_streams):
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.14_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/Users/quentinblampey/Library/Caches/pypoetry/virtualenvs/sopa-hDHgkEug-py3.10/lib/python3.10/site-packages/ipykernel/ipkernel.py", line 766, in run_closure
    _threading_Thread_run(self)
  File "/opt/homebrew/Cellar/python@3.10/3.10.14_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/quentinblampey/dev/sopa/sopa/utils/utils.py", line 381, in connect_streams
    output_stream.buffer.write(line)
AttributeError: 'OutStream' object has no attribute 'buffer'. Did you mean: '_buffers'?
```

quentinblampey:
Closing, because solved in #199

quentinblampey:
Hi @jeffquinn-msk, I have some questions regarding the nextflow pipeline you made.
I have some time these days to work on supporting nextflow, but I had some `copy` issues on Slurm due to using the same .zarr in most process. By any chance, do you have public code somewhere to see how you did that?

I have a dirty v0 here: https://github.com/quentinblampey/nf-core-sopa

quentinblampey:
Thanks @jeffquinn-msk! If I understand correctly, for now this pipeline works only for one sample (i.e. as for the Snakemake pipeline), is that right?

Also, I saw you made some docker images already, this is very convenient, thanks! I may do "official" Sopa images in the future, but for now I'll use these to prototype the pipeline!

qwwqwwq:
Right now the top-level entrypoint is for one sample yes but you can easily run multiple samples, just need compile them all in an input channel with separate names somehow (via samplesheet perhaps, which is nf-core convention), and supply that as the input channel

qwwqwwq:
Yeah the containers were a pain, specifically for baysor and stardist, hope that helps save you some trouble

quentinblampey:
Yes I think the containers will definitely save a lot of time!

PR #199, opened 2025-02-04
Baysor verbose on one patch, and propagate right error code

quentinblampey:
Alternative to #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 - Can set all `spatialdata_io.xenium` arguments in `sopa.io.xenium`
 - CLI for stardist @jeffquinn-msk (#189)
 
+### Changed
+- `sopa.io.write_report` is copying the adata to avoid modifying it (#196)
+
 ## [2.0.0] - 2025-01-20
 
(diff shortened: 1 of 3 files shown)


quentinblampey:
Thanks for your comment, that's right, I'll close this PR and go back to the other

quentinblampey:
> So one key feature I was interested in was getting the incremental output of the baysor process, so that when I look at my slurm job logs interactively I can see the ETA that baysor prints to stdout, this will only collect the stdout of the baysor process once it exits, and so wont show any incremental progress

Is it nextflow-specific? With this PR, the CLI shows the baysor logs in real time (and same for snakemake).

quentinblampey:
Sounds good. In that case, since this PR is closer to the final state that we want, I opened it back. I added your changes and added you as a "co-author commit" so that in appears that you contributed to it.

Let me know if this looks better to you!

There is one last thing I want to update: now, the error message is not as informative as before. I'll try to improve it

quentinblampey:
I also added your name to the changelogs @jeffquinn-msk.
I'm ready to merge, do you confirm all is also good on your side?

PR #203, opened 2025-02-06
Baysor parallelization per patch

Marius1311:
This adds parallelization to Baysor within each patch. See the Baysor docs here: https://kharchenkolab.github.io/Baysor/dev/configuration/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,8 +2,8 @@ fail_fast: false
 default_language_version:
     python: python3
 default_stages:
-    - commit
-    - push
+    - pre-commit
+    - pre-push
 minimum_pre_commit_version: 2.16.0
 repos:
     - repo: https://github.com/psf/black
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -176,6 +176,8 @@ rule resolve_baysor:
         smk_transcripts_temp_dir = paths.smk_transcripts_temp_dir,
     shell:
         """
+        export JULIA_NUM_THREADS={resources.cpus_per_task} # parallelize within each patch for Baysor >= v0.7
+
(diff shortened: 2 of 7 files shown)


quentinblampey:
Thanks @Marius1311 for the PR. I noticed `export` may not work on Windows (and also the `rm` command a few lines below)
I'll have to think about alternatives

quentinblampey:
Hi @Marius1311,

I think you added the `export ...` in the wrong rule. You want to use it inside `patch_segmentation_baysor`, right? Did you check that baysor is indeed using it?

Also, specifically for this rule, I think it's better to set `cpus_per_task=1` in the rule resources (see below), because some users may not use the default Sopa profiles for snakemake. If they don't use our profiles, they'll have an issue when running `patch_segmentation_baysor`, and I would prefer it to work by default. What do you think?

```sh
    resources:
        cpus_per_task=1
```

Last comment: it seems `snakemake` requires Windows users to use WSL. It may be the reason why I received no issue with the `rm` command. So, it means it should also be fine for the `export` command.

Marius1311:
Hi @quentinblampey, you're right, no idea how I managed to put that into the wrong rule; it's fixed now. I did not extensively test whether Baysor is actually using all cores now - I just followed the Baysor docs here. In my own experience, I'm now using Baysor v0.7.1 and I don't patch the images anymore (I'm using a single very large patch for the entire image) because Baysor is fast enough and memory is sufficient. I provide 10 cores and 64 GB memory (in total) for the Baysor segmentation job on a SLURM cluster. My largest images are around 30k  x 25k pixels, and Baysor takes max. 50min. Of course, this means I save time on downstream `resolve_baysor` (https://github.com/gustaveroussy/sopa/pull/197).

quentinblampey:
Thanks @Marius1311, this looks good now, merging!
This feature will be a nice addition :)

PR #208, opened 2025-02-10
Addin H-optimus-0 Model

stergioc:
This is adding the H-Optimus-0 model for tile feature extraction.


stergioc:
Ah right! The `timm` library is not available by default in sopa :|

We might need to re-write the ViT giant class here to avoid the timm dept

quentinblampey:
> We might need to re-write the ViT giant class here to avoid the timm dept

How giant is it? If it's too big, we can consider the option of only checking if `timm` is installed, and send an appropriate error log if not (something like `ERROR: Using the hoptimus0 model for inference requires installing the timm dependency, this can be done via 'pip install timm'`)

stergioc:
Sorry, I mean the class of the ViT variant *giant* that is used by the h-optimus-0. The solution of adding the `timm` library as an optional requirement in case this model is selected for inference seem to me the best tradeoff. I will update the PR :)

quentinblampey:
> Sorry, I mean the class of the ViT variant giant that is used by the h-optimus-0.

Oh I see, sorry for the misunderstanding!

Looking good now, I'm merging :)

stergioc:
Ah, maybe we need also to update the docstring for the `compute_embedings` function.

quentinblampey:
You're right, I just updated it directly on `master`

PR #209, opened 2025-02-11
Openslide backend sorting scales fix

stergioc:
In case a wsi has more than 10 levels then the image keys are ordered as: `['0', '1', '10', '2', ...]` when using the `.keys()` function. If a DataTree is initialized with this order the the multiscale image initializtion fails. This patch is sorting the scales for the DataTree generation correctly.


quentinblampey:
LGTM, thanks @stergioc!
Next time, could you please also update the CHANGELOG.md? I'll do it on master

PR #210, opened 2025-02-13
Add xenium QV threshold arg

callum-jpg:
I found myself needing a concise way to threshold transcripts based on QV when using the CLI (`sopa convert...`)


quentinblampey:
Looking good @callum-jpg, it's a nice addition, thanks!

PR #212, opened 2025-02-17
Storing patches embeddings as an `AnnData` object instead of images

quentinblampey:
@stergioc I made this PR to store patch embeddings inside an `AnnData` object instead of an image.

Notes:
- this `AnnData` object is linked to the GeoPandas dataframe containing the shape. So when plotting the `embeddings_patches` with `spatialdata_plot`, you can directly provide `color="cluster"` (where `cluster` is a column of `adata.obs`) and it will detect it. Or, for instance, you can use `color="0"` to show the first column value of the embeddings.
- the `AnnData` object also contains the patches locs in `adata.obsm["spatial"]`, meaning you can use [`sc.pl.spatial`](https://scanpy.readthedocs.io/en/1.10.x/api/generated/scanpy.pl.spatial.html) if you only care about the centroids of the patches (rather than their actual boundary).

Let me know if you like this!


stergioc:
Nice @quentinblampey :)

I will take a look at this. I expect that it will introduce some breaking changes on pipelines that are working on the image version of the features. But I think it is cleaner like this.

quentinblampey:
All the Sopa functions (like `compte_embeddings` or `cluster_embeddings`) have the same signatures and will not break, but any user-specific downstream analysis may indeed break...
I think that we don't have many H&E users yet, so it may be the best time to make this change. It may be more complicated in the future if people start using it more for H&E

stergioc:
Hi @quentinblampey,

I checked a bit the code and tried this branch on a downstream codebase I have. Indeed the API works without any issues however I got an issue when trying to update the `.attrs` field later on trying to store the feature embeddings configuration for future reference.

Here is the code I am using:
```python
sdata[embeddings_img_key].attrs["embeddings_config"] = self.config["embeddings"].copy()
sdata[embeddings_img_key].attrs["embeddings_config"].update({"level": level, "resize_factor": resize_factor})
```

I am not quite familiar with the AnnData format but would it be possible to add at the AnnData the main configuration that was used for the feature extraction (e.g., patch_width, patch_overlap, magnification, level, resize_factor, model)? Something like this can be helpful in case you want to run multiple feature extraction models and store them to the same sdata object.

sopa/patches/cluster.py
```diff
@@ -44,5 +44,5 @@ def cluster_embeddings(
         assert method in METHODS_DICT, f"Method {method} is not available. Use one of: {', '.join(METHODS_DICT.keys())}"
         method = METHODS_DICT[method]
 
-    element.obs[key_added] = method(element, **method_kwargs)
+    element.obs[key_added] = method(element.X, **method_kwargs)
```
stergioc, line 47: I had the error when I was passing as a method a `KMeans().fit_predict` so I changed the call to pass the AnnData().X here.

*** ValueError: setting an array element with a sequence. The requested array would exceed the maximum number of dimension of 64.

quentinblampey: This will then cause an issue for the Leiden method. We need to make a decision, i.e. the method either takes as input the AnnData object, or the numpy embeddings. I would be in favor of using `AnnData`, because it would allow more general operations (using variables that may be outside of `adata.X`). Still, for convenience, if you want to use KMeans, we can make add it to the available methods

What do you think @stergioc?

stergioc: Yes you are right. I removed the `.X` and added a `kmeans_clustering` method

stergioc (approved):
Looks good from my side ;)

quentinblampey:
I fixed a small typo, and it's looking good to me too, merging!

PR #218, opened 2025-02-24
Wsi reader improvements

stergioc:
### Added
- `slideio` backend for WSI is now available and allows loading among others the Olympus `.vsi` format
- `nn.DataParallel` for WSI inference now allows using multiple GPUs and larger batch_sizes

### Changed
- Removed chunking for the `io.wsi` reader (leads x10 speedup when reading tiles from the sdata)


sopa/io/reader/_openslide.py
```diff
@@ -79,6 +79,16 @@ def __init__(self, path: str, tilesize: int = 512):
         self._writeable = False
         self._erasable = False
 
+    def __contains__(self, key: str):
```
quentinblampey, line 82: Why is this needed here?

sopa/io/reader/wsi.py
```diff
@@ -36,9 +35,9 @@ def wsi(
         scale_image = DataArray(
             img[key].transpose("S", f"Y{suffix}", f"X{suffix}"),
             dims=("c", "y", "x"),
-        ).chunk(chunks)
+        )
```
quentinblampey, line 38: Do we still have chunks (since you removed `.chunk(chunks)`)? Do we sometimes have images without any chunks?

quentinblampey (commented):
Nice @stergioc! I added two comments (questions, actually)

sopa/io/reader/_wsi_reader.py
```diff
@@ -0,0 +1,249 @@
+import logging
+
+from spatialdata import SpatialData
+
+log = logging.getLogger(__name__)
+
+
+def get_reader(backend: str):
```
quentinblampey, line 8: You can maybe create a dict to map a backend name to the corresponding reader? This would avoid the numerous if/else

sopa/io/reader/_wsi_reader.py
```diff
@@ -0,0 +1,249 @@
...
+        raise NotImplementedError
+
+
+class OpenSlideReader(ReaderBase):
+    """Reader using the openslide backend."""
+
+    def __init__(self, path: str):
+        from openslide import OpenSlide
+
+        self.name = "openslide"
```
quentinblampey, line 74: You can maybe have this attribute at the class level? Since it's never updated

sopa/io/reader/wsi.py
```diff
@@ -54,6 +54,7 @@ def wsi(
     sdata = SpatialData(images={image_name: multiscale_image}, attrs={SopaAttrs.TISSUE_SEGMENTATION: image_name})
     sdata[image_name].attrs["metadata"] = slide_metadata
     sdata[image_name].attrs["backend"] = backend
+    sdata[image_name].attrs["path"] = path
```
quentinblampey, line 57: Is it saved on disk correctly if `path` is a `Path` and not a `str`?

sopa/patches/_inference.py
```diff
@@ -19,8 +21,15 @@ def __init__(
...
         magnification: int | None = None,
         device: str | None = None,
+        data_parallel: bool | list[int] = False,
     ):
         self.image, self.level, self.resize_factor = _get_extraction_parameters(image, level, magnification)
+        _backend = image.attrs.get("backend")
+        self.slide = (
+            get_reader(image.attrs.get("backend"))(image.attrs.get("path"))
+            if _backend and _backend != "zarr"
+            else get_reader("zarr")(image)
```
quentinblampey, line 31: Here we can maybe use a try/catch for non-zarr warning, which may fail e.g. if we change the path to the original image. Then, if it fails, maybe we can default to the default zarr reader and log a warning?

sopa/io/reader/_wsi_reader.py
```diff
@@ -0,0 +1,249 @@
...
+        ]
+
+    @property
+    def dimensions(self):
+        """Get the dimensions of the slide."""
+        return self.slide.get_scene(0).size
+
+
+class ZarrSlideReader(ReaderBase):
+    """Reader using the openslide backend."""
```
quentinblampey, line 234: Copy/pasted docstrings typo "openslide backend" -> "zarr backend". Btw, the name `ZarrSlideReader` is slightly confusing because if we load the image with standard readers (e.g. `dask_image`) but **do not** save on disk, then `SpatialData` doesn't use zarr

sopa/io/reader/_wsi_reader.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def __init__(self, slide: SpatialData):
```
quentinblampey, line 236: Here the type is `DataArray | DataTree` I believe?

sopa/io/reader/_wsi_reader.py
```diff
@@ -0,0 +1,249 @@
...
+        self.slide = slide
+
+    def read_region(self, location, level, size):
+        """Get a region from the slide."""
+        x_start, y_start = location
+        x_end, y_end = x_start + size[0], y_start + size[1]
+        image = next(iter(self.slide[f"scale{level}"].values()))
```
quentinblampey, line 243: This won't work with `DataArray`, but maybe it was already not working before

sopa/patches/_inference.py
```diff
@@ -31,6 +40,14 @@ def __init__(
         if self.device is not None:
             self.model.to(device)
 
+        if data_parallel:
+            if isinstance(data_parallel, list):
+                ids = data_parallel
+            else:
+                ids = list(range(torch.cuda.device_count())) if data_parallel else None
```
quentinblampey, line 47: Useless `if` because we already checked `if data_parallel` above

quentinblampey (commented):
I made a quick review @stergioc!

sopa/io/reader/_wsi_reader.py
stergioc: What about something `SpatialDataReader` ?

sopa/io/reader/wsi.py
quentinblampey: Oh it's because it's in `images.attrs`, not `slide.attrs`
Maybe we can discuss this one day together to find a solution

stergioc:
The stripping problem was due to the fact that the models were not in eval mode. I pushed an update to fix this.

The shifting problem is now much more visible:

mag 10:
image

mag 20:
image

quentinblampey:
Looking better indeed @stergioc!

The scaling issue is indeed concerning here...
It may be due to this [known issue](https://github.com/scverse/spatialdata-plot/issues/216), but here the shift is huge, so I'm afraid there is something else going wrong. What do you think?

quentinblampey:
I ran your code and was able to reproduce the issue:
image

I had a look to the patches, but they look good:

```python
>>> slide["embeddings_patches"].attrs
{'transform': {'global': Scale (x, y)
      [4. 4.]}}
```

The scale is 4, and if I apply this to the patches max coordinates:

```python
>>> centroid = slide["embeddings_patches"].centroid
>>> centroid.x.max() * 4, centroid.y.max() * 4
(79360.0, 33280.0)
```
I get around 80k and 33k, which seems consistent with the image coordinates. So, it may be an issue with `spatialdata-plot` in the end...I'll investigate more later

quentinblampey:
Nice! Do you think it would be easy/possible to add some tests for it and for all backends?

sopa/io/reader/_wsi_reader.py
```diff
@@ -0,0 +1,253 @@
...
+        raise NotImplementedError
+
+
+class OpenSlideReader(ReaderBase):
+    """Reader using the openslide backend."""
+
+    name = "openslide"
+
+    def __init__(self, path: str):
+        from openslide import OpenSlide
```
quentinblampey: Is it possible to test it openslide is installed and, if not, add information of which packages to install? As done [here](https://github.com/gustaveroussy/sopa/blob/84b787a8b9a6e66f5a158e97fa7a4426db35dafc/sopa/segmentation/methods/_cellpose.py#L122) for instance
(and same for slideio)

tests/test_patches.py
```diff
@@ -115,12 +115,17 @@ def test_patches_inference_clustering():
     assert sdata["dummy_embeddings"].shape == (4, 3)
 
     sopa.patches.compute_embeddings(sdata, "dummy", 50, magnification=100, image_key="he_image")
-    assert sdata["dummy_embeddings"].shape == (4, 3)
+    assert sdata["dummy_embeddings"].shape == (16, 3)
```
quentinblampey: Why is it now 16?

stergioc:
Hmm, that is a good idea, and it seems you've already caught something. I will take a look, it should not be a problem to retrieve any size.

stergioc:
Hey @quentinblampey,

let me try to answer your questions. 

- Regarding the transposing of the image. I tried to minimize changes to the indexing from the previous implementation to avoid any unexpected issues that might arise in parts that have already been tested. Indeed, the `read_region` expects a specific axes order. We can consider cleaning this up, but to my understanding, it should not significantly affect processing speeds.
- For the `level_downsample` scaling variable. There are cases where the magnification you are targeting is not natively available in the file, for e.g., you might have 40x, 10x, 2.5x, and you might be targeting a 5x magnification. In this case, the 2DPatches will give you bboxes at level 1 (10x), and then you will need to resize the tile to reach 5x. The `read_region` function, however, needs the coordinates at the level 0 (40x) for correct indexing. Therefore, you will need to know the downsampling factor of the level and the rescaling factor associated with that level with respect to level 0.
- It seems that the previous ResNet implementation introduced this weird behavior. I changed it to using `timm` and it works without any issue. Currently, the CI is failing because `timm` is not available during the tests, though. So I am fairly confident that the different backends/reading approaches are quite stable currently.
- Adding more models at the `wsi_backends` tests would, at this point, test mainly the models implementation and not the backends' stability. Adding clean support for `timm` models can be quite easy and useful for providing easy access to multiple models upon user authentication, but I also think that this is not relevant to the PR.  

I think just a minimal single model (e.g., `histo_ssl`) can do the trick, and we do not need to add more models to these tests.

quentinblampey:
Thanks @stergioc, it's now merged!
Btw, since you have write permissions, you can directly create a new branch from the official repository next time 🙂

PR #223, opened 2025-03-12
Add Proseg support

lguerard:
WIP to add support for proseg segmentation in sopa.


lguerard:
Here's the draft PR for proseg segmentation.

It works in theory but I'm still having issues with the transcripts.csv that gets copied to the single transcript patch folder with unequal column numbers...
I tried using the toy_dataset but that one worked. I'm not sure how to share with you anything as even the CSV that gets copied is more than 1GB... :/

lguerard:
Please also note that I haven't yet looked into the resolving step etc... So far, I'm still trying to get the segmentation to work first, but I think it's in a good direction.

I also will need to add checks to verify that proseg is installed before doing anything.

sopa/segmentation/methods/_proseg.py
```diff
@@ -0,0 +1,120 @@
...
+    get_transcripts_patches_dirs,
+)
+from .._transcripts import _check_transcript_patches, resolve
+
+log = logging.getLogger(__name__)
+
+
+def proseg(
+    sdata: SpatialData,
+    input_technology: str = "merscope",
```
quentinblampey, line 21: I think it's better to remove the `input_technology` argument, since it's already handled by Sopa. For instance, proseg uses `input_technology` to know the prior cell key, but Sopa already defines it when creating the transcript patches

sopa/patches/_factory.py
```diff
@@ -65,12 +67,22 @@ def make_transcript_patches(
...
+        sdata.points,
+        key=points_key or sdata.attrs.get(SopaAttrs.TRANSCRIPTS),
+        return_key=True,
     )
 
+    if patch_width is None:
+        # TODO: Find a way to make it have the max size
+        # patch_width = sdata.attrs.get(SopaAttrs.PATCH_WIDTH)
+
+        patch_width = max(get_spatial_image(sdata).shape)
```
quentinblampey, line 84: I think it would be better to directly support `patch_width: None` in the `Patches2D` and `Patches1D` classes. I can do that if you want

quentinblampey (commented):
Thanks @lguerard, I think it's a very good start! I made some comments, but I think it's already relatively close to the final result.
I'll probably find some time to help you finishing it

sopa/patches/_factory.py
quentinblampey: I added a `patch_width = patch_width or float("inf")` in `Patches2D` to support it

sopa/segmentation/methods/_proseg.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+
+    patch_dir = get_transcripts_patches_dirs(sdata)
+    if len(patch_dir) > 1:
+        raise IndexError(
+            "Proseg is fast enough to work on a single patch. Support for multiple patches is not yet implemented. Rerun the transcript patches step with patch_width=None."
+        )
+
+    settings.parallelization_backend = None
+    settings._run_with_backend(
```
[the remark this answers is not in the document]
lguerard: That's true, I just copied it from the `Baysor` file but wanted to check if it would still work without that..

sopa/segmentation/methods/_proseg.py
```diff
@@ -0,0 +1,120 @@
...
+        [partial(proseg_run, patch_dir) for patch_dir in patch_dir]
+    )
+
+    # resolve(sdata, gene_column, min_area=min_area, key_added=key_added)
```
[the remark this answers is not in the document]
lguerard: Ok I'll look into it. Is the transcript mapping to the cell done automatically when processing ?

sopa/segmentation/methods/_proseg.py
lguerard: That's true. I was working on improving the default config for proseg before pushing, so still looking into it.

sopa/patches/_factory.py
lguerard: Thanks a lot !

sopa/segmentation/methods/_proseg.py
lguerard: Actually, proseg needs this input technology to figure out the CSV column names from the transcripts.csv as you can see [here](https://github.com/dcjones/proseg/blob/704f8f211786f9b23ed0b1f8ae47af02b5c6e106/src/main.rs#L377).

Is sopa changing these names to something common regardless of the input ? If so, we can indeed discard it, otherwise it would still be needed.

quentinblampey: Yes, Sopa will change the coordinates to x / y / z for all technologies

sopa/segmentation/methods/_proseg.py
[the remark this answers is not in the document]
quentinblampey: It's now done

lguerard:
Hi @quentinblampey 

Thanks a lot for your very fast and efficient work, so happy to see this moving so fast 😄 

With the most recent changes I tried it twice and got some errors:
* Tried on real data, and I'm still getting an error with the transcripts.csv file not having equal lines `called `Result::unwrap()` on an `Err` value: Error(UnequalLengths { pos: Some(Position { byte: 80492597, line: 1035293, record: 1035292 }), expected_len: 11, len: 8 })`
* Tried it on the toy_dataset but after running cellpose (I was just curious to see how it compares to a full pipeline), and it failed with `Column 'z' not found in CSV file`. I checked the CSV and the Z column is called `z_`, is that expected ?

quentinblampey:
> Tried it on the toy_dataset but after running cellpose (I was just curious to see how it compares to a full pipeline), and it failed with Column 'z' not found in CSV file. I checked the CSV and the Z column is called z_, is that expected ?

Are you sure you pulled the changes? I made a little update to fix this.
For some context: certain datasets/technologies may not have a `z` column, so if there is no `z` column I create a fake column with zeroes before running proseg. For the dummy dataset, the `z_` column was used before for testing purposes, but I also changed it as it was not useful.

> Tried on real data, and I'm still getting an error with the transcripts.csv file not having equal lines

I see, this is probably still the same issue you had with Baysor (i.e. this is likely not related to proseg though).
Could you please open a new issue with precise details of your OS + full dependencies versions? Do you use conda or another package manager?
Since it doesn't happen on the small toy dataset, maybe it only happens on larger datasets. Can you try using a higher `length` when creating the toy dataset to see if it happens? 

Sorry for not being able to fix this - I really can't reproduce the issue, despite running Sopa for many different machines, on different OS, and different installation approaches.

lguerard:
I pulled everything and installed sopa as editable so I thought it would pick up everything, but maybe not. I'll recreate the environment and try again, sorry for the confusion.



> Sorry for not being able to fix this - I really can't reproduce the issue, despite running Sopa for many different machines, on different OS, and different installation approaches.

Not at all your fault, I know you're trying as best you can ! I'll make an issue and post all the info there. I'll also try the `length` option and let you know.

quentinblampey:
Great, no problem! Normally, it should be fine if it's in editable mode. Maybe check that you use the branch `proseg` on your fork and that your editable installation points to your fork?

If it works for you on the toy dataset, I'll merge it, and we'll fix the transcript issue separately.

lguerard:
It worked indeed ! Here's the result using Napari and some settings.

image

There seems to still be some issues with proseg as you can see here : 
image

I know this is not sopa's fault, and these could potentially be fixed with changing some proseg settings. But that brought the question of support for additional input options, is that something feasible in your opinion ?

quentinblampey (approved):

quentinblampey:
Perfect, thanks, I'm merging now!

quentinblampey:
Hi @lguerard, I didn't know that it was possible with proseg. So, yes, I would be happy to review your PR!

Issue #180, opened 2025-01-21
[Feature] Clarify how to get the CosMX input


quentinblampey:
Some users mentioned it was difficult to get the right Sopa's input from AtomX. This should be clarified with Nanostring

professor-sagittarius:
Hi Quentin. I've been following the various threads about AtoMx output. Just wanted to make sure you were aware of this documentation that contains the file structure that is exported by the current version of AtoMx (1.3.2).

[CosMxFlatFileExportSetup.docx](https://github.com/user-attachments/files/19116578/CosMxFlatFileExportSetup.docx)
[CosMxDAExportSetup.docx](https://github.com/user-attachments/files/19116577/CosMxDAExportSetup.docx)

Briefly, most of the required CosMx files are exported via a FlatFileExport.R script, while the Morphology2D TIFs are now exported via a separate DAExport.R script.

Notably, as several users mentioned, Morphology_ChannelID_Dictionary.txt is no longer exported. I have some CosMx data from about a year ago, and this dictionary file is not present there either.

quentinblampey:
Thanks @professor-sagittarius for your comment, this will help!
Since `Morphology_ChannelID_Dictionary.txt` is no longer exported, do you know how to access the channels names now? And do you know where I could find up-to-date public CosMX data that I could use to debug that? Sorry we don't have a CosMX anymore in our lab so it's more difficult for me to work on CosMX data...

professor-sagittarius:
@quentinblampey, Bruker has a [whole-transcriptome pancreas dataset](https://smi-public.objects.liquidweb.services/cosmx-wtx/Pancreas-CosMx-WTx-FlatFiles.zip) available for download, and the file structure matches what's exported by AtoMx. 

I was looking through the files, and I think the best way to get morphology channel names is from the header of `metadata_file.csv`. Here are the relevant column names from the public dataset above:
`Mean.PanCK,Max.PanCK,Mean.CD68_CK8_18,Max.CD68_CK8_18,Mean.CD298_B2M,Max.CD298_B2M,Mean.CD45,Max.CD45,Mean.DAPI,Max.DAPI`
As proof that the columns change depending on input channels, here's a header from some of my data that used the Neuroscience segmentation kit:
`Mean.Histone,Max.Histone,Mean.CK8.18,Max.CK8.18,Mean.rRNA,Max.rRNA,Mean.GFAP,Max.GFAP,Mean.DAPI,Max.DAPI`

[...]

PR #228, opened 2025-03-19
Corrected channel order for channel name input from TIF metadata in CosMx reader

professor-sagittarius:
Adjusted indexing of the CosMx reader's TIF metadata scraper to ensure channels are output in the same order as they were when `Morphology_ChannelID_Dictionary.txt` was used to get channel names. #180

### DIFF
--- a/sopa/io/reader/cosmx.py
+++ b/sopa/io/reader/cosmx.py
@@ -268,7 +268,10 @@ def _cosmx_morphology_coords(images_dir: Path) -> list[str]:
     with tifffile.TiffFile(images_paths[0]) as tif:
         description = tif.pages[0].description
         substrings = re.findall(r'"BiologicalTarget": "(.*?)",', description)
-        return substrings
+        channels = re.findall(r'"ChannelId": "(.*?)",', description)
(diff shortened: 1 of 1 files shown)


quentinblampey:
Hi @professor-sagittarius, thanks for the PR!
Do you know if there is a Nanostring documentation regarding the right channel order?
Intuitively, I would have used `stdorder = ["U", "B", "G", "Y", "R"]`, i.e. sorted by wavelength, but maybe Nanostring did it differently. I just wanted to double check with you

professor-sagittarius:
Hi @quentinblampey, great questions. I was using the channel order from `Morphology_ChannelID_Dictionary.txt`:

> ChannelID       BiologicalTarget
> Channel-B       Histone
> Channel-G       CK8/18
> Channel-Y       rRNA
> Channel-R       GFAP
> Channel-U       DNA

However, you are right that ideally there would be some way to retrieve this information without hard-wiring it into the sopa codebase.

Further investigation led me to find that the TIF metadata has the line `"ChannelOrder": "BGYRU",`, so I have added a commit to re-index the channels by this information instead.

PR #233, opened 2025-03-25
Work on making the prior segmentation optional for Proseg

lguerard:


lguerard:
Hmm actually not that easy, as the `cell-id-column` needs to be set, therefore might need a prior... I'll dig a bit more and we can keep this as draft in the meantime.

quentinblampey:
Yes this was what I observed when working on your previous PR. This is why I assumed a prior was always needed, but it may be wrong, let me know if you find a workaround

quentinblampey:
I'll close the issue because, as said above, Proseg probably can't be run without a prior. Don't hesitate to re-open if you found out I was wrong!

Issue #232, opened 2025-03-24
[Feature] Remove `cv2` dependency


quentinblampey:
`open-cv` is one of the heavy dependencies of `sopa`. It causes some issues when making a docker image, as it requires to apt-get `ffmpeg libsm6 libxext6`, which makes the image much bigger. 

I created a branch `nocv2` where I removed the usage of `cv2` in most functions. Still missing:
- minor adjustment to reproduce exactly the behavior we had with OpenCV (i.e., fix tests)
- update the Otsu tissue segmentation

@stergioc do you think it would be easy to update the tissue segmentation to remove `cv2`? E.g., using [`skimage`](https://scikit-image.org/) instead, which is a dependency of `spatialdata` anyway

stergioc:
Thats a good idea. Yes I will take a look at this the next days it should be easy to do. 👍

stergioc:
Hi @quentinblampey, I just pushed this 73917f805d97ed8a315278d1afd88240646e7155, where I removed the `cv2` dept from the tissue segmentation functions. I manually checked the different steps for few WSIs but I did not do extensive tests. I hope it helps :)

quentinblampey:
Awesome @stergioc! I'll have a look and try to run more tests

We'll need many tests to ensure this doesn't create regression.
I think we could add this version in `sopa==2.1.0` (not expected to be released before a few months), since it's a too big change to be pushed as a minor release

quentinblampey:
Hi @stergioc, I had a look to your updates, it looks very good. I added a script in the tests to check how different the new implementation is from the old one. The results are very similar, and many steps are identical.

I only changed the part where you convert the labels to the polygons, using the `_vectorize_mask` function that I wrote in `spatialdata` (I had some weird shapes with your implementation).

There is still a minor issue, that we already had before using `cv2`, but still now. When we have a mask with an interior, it is not kept when converted to a Polygon. For instance, for this mask:

![Image](https://github.com/user-attachments/assets/b540baac-9f51-4740-b0b0-98cb1eeb9550)

It is converted to a square:

![Image](https://github.com/user-attachments/assets/7f91134d-8e67-423e-b191-73f9deeec40f)

[...]

PR #239, opened 2025-04-14
Remove `open-cv` dependency

quentinblampey:


stergioc:
commit [3c176a5](https://github.com/gustaveroussy/sopa/pull/239/commits/3c176a54b316287fd6d9c73923e2a8be37aa7b23) should resolve https://github.com/gustaveroussy/sopa/issues/232

quentinblampey:
Thanks @stergioc, it works beautifully! The method is very clean, I'm convinced it should be very reliable.
I pushed some commits to allow holes only for tissue segmentation (for cell segmentation, holes are not desired)

I think the PR is now ready, I'll merge it before releasing `sopa==2.1.0` which will be a big release. But I still need to add some extra features that I want to add in `2.1.0`, I aim to release it in April

quentinblampey:
I realize that my `rasterize` method doesn't handle holes yet. So although we have the right shapes, when rasterizing them for cell segmentation, we lose the holes again 😅
(This rasterization is needed to apply the tissue segmentation as a mask on the image before segmentation)

This is a different issue, so I'll open a new issue and try to correct it later

quentinblampey:
If you want @stergioc I can move the internal function `_region_props_to_multipolygon` with the "holes support" to `SpatialData`, and then we'll call it inside Sopa for more consistency. What do you think?

PR #242, opened 2025-04-18
Docker images release job

Clemsazert:
Create a job to release docker images to dockerhub registry.

To do before merge :
[] Create a config variable DOCKERHUB_USERNAME with the dockerhub account username value
[] Create a repository secret variable DOCKERHUB_TOKEN with the token from dockerhub

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,11 +86,14 @@ jobs:
         extra-name: ["", "baysor", "cellpose", "proseg", "stardist"]
 
     env:
-      EXTRA_NAME_SUFFIX: ${{ matrix.extra-name == "" && "" || format('-{0}', matrix.extra-name) }}
-      # Extract the 'v' character from the tag to only keep the version numbers.
-      STRIPPED_TAG_VERSION: ${{ github.ref_name == startsWith(github.ref_name, 'v') && github.ref_name[1:] || github.ref_name }}
+      EXTRA_NAME_SUFFIX: ${{ matrix.extra-name == '' && '' || format('-{0}', matrix.extra-name) }}
 
     steps:
+      - name: Export tag version
+        run: |
+          TAG="${GITHUB_REF#refs/tags/v}"
+          echo "VERSION=$TAG" >> $GITHUB_ENV
+
       - name: Login to Docker Hub
         uses: docker/login-action@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -75,3 +75,33 @@ jobs:
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          push: true
+          platform: linux/amd64,linux/arm64
+          tags: ${{ vars.DOCKERHUB_USERNAME }}/sopa:${{ github.ref_name }}${{ env.EXTRA_NAME_SUFFIX }}
```
quentinblampey, line 106: In that case, `github.ref_name` corresponds to the `tag` name, right? Is it possible to remove the `v` prefix, to only have e.g. `2.0.5` instead of `v2.0.5`?
In my [Docker hub](https://hub.docker.com/repository/docker/quentinblampey/sopa/tags), I wasn't adding the `v`, but I don't know what's the convention for Docker images

.github/workflows/ci.yml
```diff
@@ -75,3 +75,33 @@ jobs:
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          push: true
+          platform: linux/amd64,linux/arm64
```
quentinblampey, line 105: We can't install Baysor via the binaries on `arm` - see their releases [here](https://github.com/kharchenkolab/Baysor/releases) (for `arm`, they recommend installing Baysor in Julia). Could we skip the `sopa-baysor` release specifically for arm, or have a different Dockerfile?

quentinblampey:
Thanks @Clemsazert! I'm going to add the secrets

.github/workflows/ci.yml
Clemsazert: Sure we can skip linux/arm64 for this one.
Btw did you try to build the other images on your local for the linux/arm64 platform ? Could be good to git it a shot before building the automation.

quentinblampey: Yes, good point, I can try that (probably next week)!

quentinblampey (approved):

quentinblampey:
I pushed some changes @Clemsazert:
- I used single quotes instead of double quotes (else the CI raised a syntax error)
- I removed the arm support, I'll see later if I add it back
- I'm not sure if it was possible to strip the version tag as you did. I did it differently, it's ugly but I couldn't find an easier way...Let me know if you know how to improve it

quentinblampey:
I'll merge it to release a new version, and we can make another PR later if the build fails

PR #262, opened 2025-05-19
Add support for other conda env

lguerard:
Probably not the most pretty way, but this allows to have the conda_env as a variable.

Could be improved by adding the environment name as an argument to snakemake, but I don't know enough about it...


lguerard:
That seems way cleaner than my way indeed!

Do you want to do it or should i do it?

quentinblampey:
If you want to do it then go for it :)

PR #263, opened 2025-05-19
Fix snakemake aggregate call

lguerard:
Since the call got updated, this update the snakemake recipe as well.

### DIFF
--- a/workflow/config/merscope/cellpose.yaml
+++ b/workflow/config/merscope/cellpose.yaml
@@ -18,7 +18,7 @@ segmentation:
 
 aggregate:
   aggregate_channels: true
-  gene_column: "gene"
+  aggregate_genes: true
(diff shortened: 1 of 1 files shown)


quentinblampey:
Hi @lguerard, thanks for the PR. Actually, although deprecated, the CLI still supports `gene_column` for backward compatibility. That is, if `gene_column` is provided, it will send a warning and use `aggregate_genes = True` instead.

So, both works and have the same behavior.

I think I'll update all the configs to use `aggregate_genes` in the `2.1.0` release, but I'll keep the backward compatibility since some users may still use old config files.

I'll close the issue since it is specific to one config file, but I will keep in mind to update all the config files soon.

lguerard:
Hi @quentinblampey, that makes sense, thanks! And sorry for the extra work 😅

PR #264, opened 2025-05-19
Start adding support for Cellpose4

lguerard:
Not tested yet with other models than `cpsam` but works for that model.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-cellpose = ["cellpose>=3.0.5, <4.0.0", "torch>=2.0.0, !=2.0.1, !=2.1.0"]
+cellpose = ["cellpose>=3.0.5", "torch>=2.0.0, !=2.0.1, !=2.1.0"]
 stardist = ["tensorflow>=2.18.0", "stardist>=0.9.1", "numpy<2.0.0"]
 baysor = ["toml>=0.10.2", "loompy>=3.0.7"]
 wsi = [
--- a/sopa/segmentation/methods/_cellpose.py
+++ b/sopa/segmentation/methods/_cellpose.py
@@ -126,7 +126,7 @@ def cellpose_patch(
         if pretrained_model:
             model = models.CellposeModel(pretrained_model=pretrained_model, **cellpose_model_kwargs)
         else:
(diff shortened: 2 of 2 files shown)


quentinblampey:
Thanks for your PR @lguerard, it's a very good start. Note that we'll need backward compatibility, i.e., if users are still using `cellpose<4`, it has to also work.
I'll try to test it soon

lguerard:
Yes and also the other models than Cellpose has. I can work on it a bit more in 2 weeks and make new pushes!

I'm also planning on working on adding an option to dilate labels, I'm working on adding it only for cellpose now, but should I try to make it compatible for all methods?

quentinblampey:
We already have cell expansion, which is (I believe) the same as what you call dilatation, is it correct? Or do you talk about something else? See `expand_radius_ratio` in [sopa.aggregate](https://gustaveroussy.github.io/sopa/api/aggregation/#sopa.aggregate)

lguerard:
No, that's exactly the same thing! But does this modify the labels or is it "just" to assign the points to cells?

quentinblampey:
It doesn't update the shapes themselves, but you can update it via the `expand_radius` function:

```python
from sopa.segmentation.shapes import expand_radius

new_shapes = expand_radius(sdata.shapes["cellpose_boundaries"], expand_radius_ratio=1)

# then you can add them to sdata
sdata["expanded_shapes"] = new_shapes
```

lguerard:
Oh that's amazing, thanks a lot!

I'll work on cellpose back compatibility then!

lguerard:
The `expand_radius` doesn't care about labels overlapping though, does it ? 

I was more looking for something where it would expand until the radius or until it finds a cell. Is that existing ?

quentinblampey:
Not yet in Sopa, but I want to add it soon, likely in `sopa==2.1.0`

quentinblampey:
Hi @lguerard, I had a look at the new version of cellpose, and it will be added to `sopa==2.1.0`. Actually, the way I wanted to change it was slightly different from what you did, so I updated it directly myself - sorry for the time spent on that PR.
Your code helped me, though, so I'll mention your help in the release and CHANGELOG!

quentinblampey:
Btw, in `sopa==2.1.0`, the `expand_radius` function will be located at `sopa.shapes.expand_radius`, and will include a `no_overlap: bool` argument

Issue #241, opened 2025-04-16
[Feature]Add a table_key argument to sopa.aggregation.aggregate_bins in order to save nuclei and cells


NadineBestard:
**Is your feature request related to a problem? Please describe.**
When aggregating the bins into cells with sopa.aggregate, the default expansion is 1, which corresponds to only aggregating the bins that fall into the nuclei. To expand to the surrounding cytoplasm, my understanding would be to use an expansion argument of around 4 (at least that is the default from bin2cell). So far, the function saves the results into "table" meaning we have to choose one or the other. 
It may be useful to add an argument in order to choose a different name for the table, and be able to call one of them "table_nuclei" and another something else such as "table_expansion_4" or "table_cells" 

**Describe the solution you'd like**

`sopa.aggregate(sdata, aggregate_channels=False, expand_radius_ratio=1, image_key="hires_image", **table_key = "table_nuclei"**)`

**Extra**
It would also be nice if it were possible to visualise the expansion as a new shape. I see you suggested an option in https://github.com/gustaveroussy/sopa/issues/130, but I am not sure how to translate this to the mean*expansion_ratio, which is used in this case for the expansion

quentinblampey:
Your code looks good @NadineBestard! I think we can slightly update it to make it smoother:
- You don't need to precise `shapes` in `sdata.shapes['nuclei_boundaries']`, you can use `sdata['nuclei_boundaries']` directly. Both works, choose the solution that feels better for you.
- I first create the two shapes layer, and then I aggregate with `expand_radius_ratio=0`. It'll result in the same, but this way you avoid duplicating the `expand_radius_ratio=2` argument, which could lead to issues in case you update one but not the other. Note that I provide `shapes_key` to `sopa.aggregate`.

```python
# save new shapes layers
sdata['nuclei_boundaries'] = sdata['stardist_boundaries'].copy() 
sdata['cell_boundaries'] = expand_radius(sdata['nuclei_boundaries'], 2) 

# aggregate for both shapes layers
sopa.aggregate(sdata, aggregate_channels=False, expand_radius_ratio=0, image_key="cf_hires_image", shapes_key="nuclei_boundaries", key_added = "table_nuclei")
sopa.aggregate(sdata, aggregate_channels=False, expand_radius_ratio=0, image_key="cf_hires_image", shapes_key="cell_boundaries", key_added = "table_cell")
```

[...]

PR #267, opened 2025-05-27
Allow to expand cells radius without touching

quentinblampey:
Adding a function `remove_overlap` to remove overlaps between polygons. Experimental feature.


quentinblampey:
Also, note that there are two possibilities:
1. What you did, i.e. expanding the cells without overlap, and then aggregating the bins
2. Expanding the cells during the aggregation, and then ensure that each bin is mapped to only one cell, based on gene expression profile. This is different from option 1 because it uses the bins expression and not just the shapes.

If you want to try option 2, then you need to provide `no_overlap` to `sopa.aggregate` directly:
```python
sopa.aggregate(sdata, aggregate_channels=False, expand_radius_ratio=2, image_key="cf_hires_image", shapes_key="nuclei_boundaries", key_added = "table_nuclei", no_overlap=True)
```

I think that it's a little bit confusing to have these two options. Maybe I should rename use a different variable name in `sopa.aggregate`, like `unique_mapping: bool`, what do you think?
Your feedback is precious, I need it to make sure the version `2.1.0` will be stable and easy to understand :)

NadineBestard:
Thank you! It's so nice you are already implementing to look at the GEX when assigning the bins to different cells :)
I noticed there was the `no_overlap` option into `sopa.aggregate`, but indeed I did not notice it was not doing exactly the same thing.
I think most people would use the aggregate directly, as you intend, as that is what is shown in the tutorial and is the logical workflow. And `no_overlap` is a good argument name. 
I followed the workflow 'creating shapes -> aggregating the bins in those shapes' as this is more or less what you suggested in https://github.com/gustaveroussy/sopa/issues/241#issuecomment-2900191919 when I was wondering how to add the expanded cell shape. 
If using your proposed option, with `expand_radius_ratio` and `no_overlap` passed directly into the aggregation, then how would you add the resulting shapes into the data? If `no_overlap` from `expand_radius` is different from the one in `aggregate` the trick used in https://github.com/gustaveroussy/sopa/issues/241#issuecomment-2900191919 would not work. 

I'll test with the main branch on Monday or so!

NadineBestard:
Still the same error with the main branch. Here is the complete log
</details>

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[10], line 3
      1 # aggregate for both shape layers
      2 sopa.aggregate(sdata, aggregate_channels=False, expand_radius_ratio=0, image_key="cf_hires_image", shapes_key="nuclei_boundaries", key_added = "table_nuclei")
----> 3 sopa.aggregate(sdata, aggregate_channels=False, expand_radius_ratio=0, image_key="cf_hires_image", shapes_key="cell_boundaries", key_added = "table_cell") 

File ~/CRM_SingleCell/Spatial/VisiumHD/src/sopa/sopa/aggregation/aggregation.py:79, in aggregate(sdata, aggregate_genes, aggregate_channels, image_key, points_key, gene_column, shapes_key, bins_key, expand_radius_ratio, min_transcripts, min_intensity_ratio, no_overlap, key_added)
     76     key_added = f"{aggr.shapes_key}_{SopaKeys.TABLE}"
     77     log.info(f"key_added is None, saving the table as '{key_added}' by default.")
---> 79 aggr.compute_table(
     80     aggregate_genes=aggregate_genes,
     81     aggregate_channels=aggregate_channels,
     82     expand_radius_ratio=expand_radius_ratio,
     83     min_transcripts=min_transcripts,
     84     min_intensity_ratio=min_intensity_ratio,
     85     gene_column=gene_column,
     86     no_overlap=no_overlap,
     87     key_added=key_added,
     88 )

File ~/CRM_SingleCell/Spatial/VisiumHD/src/sopa/sopa/aggregation/aggregation.py:221, in Aggregator.compute_table(self, aggregate_genes, aggregate_channels, gene_column, expand_radius_ratio, min_transcripts, min_intensity_ratio, no_overlap, key_added)
    214 self.sdata.shapes[self.shapes_key] = self.geo_df
    216 self.table.uns[SopaKeys.UNS_KEY] = {
    217     SopaKeys.UNS_HAS_TRANSCRIPTS: aggregate_genes,
    218     SopaKeys.UNS_HAS_INTENSITIES: aggregate_channels,
    219 }
--> 221 self.add_standardized_table(key_added)

File ~/CRM_SingleCell/Spatial/VisiumHD/src/sopa/sopa/aggregation/aggregation.py:224, in Aggregator.add_standardized_table(self, key_added)
    223 def add_standardized_table(self, key_added: str):
--> 224     add_standardized_table(
    225         self.sdata, self.table, self.geo_df, self.shapes_key, key_added, image_key=self.image_key
    226     )

File ~/CRM_SingleCell/Spatial/VisiumHD/src/sopa/sopa/aggregation/aggregation.py:242, in add_standardized_table(sdata, table, geo_df, shapes_key, table_key, image_key)
    238 geo_df.index = list(table.obs_names)
    240 add_spatial_element(sdata, shapes_key, geo_df)
--> 242 table.obsm["spatial"] = np.array([[centroid.x, centroid.y] for centroid in geo_df.centroid])
    243 table.obs[SopaKeys.REGION_KEY] = pd.Series(shapes_key, index=table.obs_names, dtype="category")
    244 table.obs[SopaKeys.SLIDE_KEY] = pd.Series(image_key or "None", index=table.obs_names, dtype="category")

File ~/CRM_SingleCell/Spatial/VisiumHD/src/sopa/sopa/aggregation/aggregation.py:242, in <listcomp>(.0)
    238 geo_df.index = list(table.obs_names)
    240 add_spatial_element(sdata, shapes_key, geo_df)
--> 242 table.obsm["spatial"] = np.array([[centroid.x, centroid.y] for centroid in geo_df.centroid])
    243 table.obs[SopaKeys.REGION_KEY] = pd.Series(shapes_key, index=table.obs_names, dtype="category")
    244 table.obs[SopaKeys.SLIDE_KEY] = pd.Series(image_key or "None", index=table.obs_names, dtype="category")

AttributeError: 'NoneType' object has no attribute 'x'
```

I tested (option2) and that works though! So it would just be a matter of sorting how to add the cell shapes back to the object in order to visualise if the expansion looks reasonable.

quentinblampey:
Great, now the error makes more sense. Maybe there is an invalid geometry inside `sdata['cell_boundaries']`. Can you run the following lines for me? It will help me debugging.

```python
print(len(sdata["cell_boundaries"]))
print(sdata["cell_boundaries"].geometry.map(lambda x: x.__class__.__name__).value_counts())
print(sdata["cell_boundaries"].centroid.map(lambda x: x.__class__.__name__).value_counts())
```

>  how would you add the resulting shapes into the data?

The problem with this approach is that the bins associated to one cell are not necessarily always next to each other. In other words, the cell would have a very weird shape, with multiple small holes in between...

PR #282, opened 2025-07-04
Feat/min patch size parameter

### LINKED ISSUE #280: Cellpose segmentation fails with OpenCV resize error on very small patches
Hey @quentinblampey! 👋

Problem
When running Cellpose segmentation through the Snakemake pipeline, I encounter an OpenCV resize error when processing very small image patches (width or height < 5 pixels). This causes the entire segmentation job to fail.

Environment:

sopa 2.1
Python 3.12
Snakemake pipeline
Error:

```
cv2.error: OpenCV(4.10.0) /io/opencv/modules/imgproc/src/resize.cpp:4152: error: (-215:Assertion failed) !ssize.empty() in function 'resize'
```

Root Cause
Edge patches in the patching process can sometimes have dimensions smaller than what OpenCV's resize function can handle, particularly when patches have 0 width or height due to irregular tissue boundaries or very small tissue fragments.

Solution
I've implemented a minimum patch size check in StainingSegmentation._run_patch():

```Python
def _run_patch(self, patch: Polygon) -> gpd.GeoDataFrame:
    bounds = [int(x) for x in patch.bounds]
    
    # Check minimum patch size to avoid resize errors
    patch_width = bounds[2] - bounds[0]
    patch_height = bounds[3] - bounds[1]
    
    if patch_width < self.min_patch_size or patch_height < self.min_patch_size:
        log.warning("Skipping patch with dimensions %dx%d (smaller than min_patch_size=%d)",
                   patch_width, patch_height, self.min_patch_size)
        return gpd.GeoDataFrame(columns=['geometry'], geometry=[])
    
    # ... rest of segmentation logic
```

This prevents the pipeline from crashing while providing transparency about which patches are being skipped. The min_patch_size parameter is configurable via CLI (default: 5 pixels).

Happy to submit a PR if this approach looks good!

Marius1311:
Add a min patch size parameter (defaults to 10 pixels) to discard very small patches that could potentially break the pipeline. 

closes #280

### DIFF
--- a/sopa/cli/segmentation.py
+++ b/sopa/cli/segmentation.py
@@ -34,7 +34,7 @@ def cellpose(
         1, help="Parameter for scipy gaussian_filter (applied before running cellpose)"
     ),
     min_patch_size: int = typer.Option(
-        5,
+        10,
         help="Minimum patch size (in pixels) for both width and height. Patches smaller than this will be skipped to avoid segmentation errors.",
     ),
     patch_index: int = typer.Option(
@@ -106,7 +106,7 @@ def stardist(
         help="Parameter for scipy gaussian_filter (applied before running the segmentation method)",
     ),
     min_patch_size: int = typer.Option(
-        5,
+        10,
         help="Minimum patch size (in pixels) for both width and height. Patches smaller than this will be skipped to avoid segmentation errors.",
     ),
     patch_index: int = typer.Option(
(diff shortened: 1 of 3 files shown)


quentinblampey:
Hi @Marius1311,

Sorry for the delay, I'll merge this soon!

I just have one question: do you think the min_patch_size argument is very useful for the CLI? I'm assuming 10 is a very good value for a minimal patch size, and I think we could use it as a constant to avoid adding arguments to the CLI (which already has many, many arguments).
What do you think? Do you see any use case where 10 would be a bad value?

Marius1311:
sure, I think 10 would be a good value!

PR #327, opened 2025-09-30
Fix typo in FAQ title

mashehu:


quentinblampey:
Good catch, thanks @mashehu!

Issue #342, opened 2025-10-28
[Feature] Nightly CI runs


quentinblampey:
Right now, we run the CI on every PR or version release. It would be great to run the pipelines at defined time steps (for instance, once every 2 or 3 days) to check if anything broke. Typically, some dependency updates may break the package installation at any time.

PR #344, opened 2025-10-28
Add nightly CI runs + wait for PyPi propagation before docker image build

### LINKED ISSUE #341: [Feature] Ping PyPI before running the Docker step in the actions
In the CI, when creating a new tag, we release the new version on PyPI, and then install this exact version on Docker images that we publish.

If the Docker step starts too quickly, the version may not yet be ready to be installed, and it may fail. Re-running the CI manually 5 minutes after is an easy fix, but we could find a better solution.

One potential solution would be to define a rule between the PyPI release and the Docker images: we could do a for loop that pings the new Sopa version on the PyPI server, and succeed only once it can detect the new version.

clementpiat:
Fixes #341 and #342
CI scheduled at 00:00 on Monday, Thursday, and Saturday.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
           python-version: "3.12"
 
       - name: Install sopa with pip
-        run: pip install -e . && pip install pytest-cov poethepoet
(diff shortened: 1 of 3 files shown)


quentinblampey (changes requested):
Thanks @clementpiat, I made some comments!

PR #411, opened 2026-04-07
feat: add density map support to Xenium Explorer transcript writer

jackytamkc:
## Summary

While making paper figures I noticed that the density map toggle in Xenium Explorer v4.0+ no longer works with SOPA-converted data, though it worked in Explorer <4.0. Investigation revealed that `write_transcripts()` in `sopa/io/explorer/points.py` only writes the `grids/` group, but native Xenium `transcripts.zarr.zip` files include additional arrays that Explorer v4+ requires for the density map.

### What's missing vs native Xenium format

| Group | Purpose |
|---|---|
| `density/gene/` | Pre-computed transcript density as CSR sparse matrix (10µm bins) |
| `density/codeword/` | Per-codeword density (same CSR format) |
| `gene_category` | Boolean matrix (n_genes × 9) classifying genes as gene/neg_control/blank/etc |
| `codeword_category` | Boolean matrix (n_codewords × 9) for codewords |
| `metrics_density` | Float32 3D array (y_bins × x_bins × 4) of QC density metrics (20µm bins) |

### Changes

- **`sopa/io/explorer/points.py`**: Added `_write_density()` and `_classify_gene_names()` helper functions. `_write_density()` is called at the end of `write_transcripts()` to write all missing arrays into the zarr group.
- **`sopa/io/explorer/_constants.py`**: Added `DENSITY_GRID_SIZE` (10µm) and `METRICS_DENSITY_SPACING` (20µm) constants.

No version bump, no new dependencies (uses `scipy.sparse` which is already a SOPA dependency), no changes to `converter.py` or any other file.

### How it was verified

- Downloaded native Xenium tiny datasets from 10x Genomics (XOA 3.0 Mouse Ileum, XOA 4.0 Human Ovary) to map the exact zarr structure
- Generated a SOPA explorer output from an in-house CosMx dataset, then patched in the density arrays
- Confirmed the density map toggle works correctly in Xenium Explorer with the patched output

Claude was used to help with the coding part.

## Test plan

- [x] Open a SOPA-generated `.explorer` output in Xenium Explorer v4.0+ and verify the density map toggle works
- [x] Verify transcript points still display correctly (no regression)
- [x] Test with Xenium, CosMx, and MERSCOPE data if available

### DIFF
--- a/sopa/io/explorer/points.py
+++ b/sopa/io/explorer/points.py
@@ -17,6 +17,133 @@ def subsample_indices(indices: np.ndarray, factor: int = 4) -> np.ndarray:
     return np.random.choice(indices, len(indices) // factor, replace=False)
 
 
+def _classify_gene_names(names: list[str], n_cols: int = 9) -> np.ndarray:
+    """Build a boolean category matrix classifying gene/codeword names.
+
+    Columns follow the native Xenium format:
+        0: is_gene, 1: is_negative_control_probe, 2: is_negative_control_codeword,
+        3: is_deprecated, 4: is_predicted_false_positive, 5: is_unassigned_codeword,
+        6: is_blankcodeword, 7: is_anti_sense, 8: is_genomic_control
+    """
+    cats = np.zeros((len(names), n_cols), dtype=bool)
+    for i, name in enumerate(names):
+        name_lower = name.lower()
+        if "negcontrol" in name_lower or "negprobe" in name_lower:
+            cats[i, 1] = True
+        elif "blank" in name_lower:
+            cats[i, 6] = True
+        elif "systemcontrol" in name_lower or "falsecode" in name_lower:
+            cats[i, 3] = True
+        else:
+            cats[i, 0] = True
+    return cats
+
+
+def _write_density(
+    root: zarr.Group,
+    xy: np.ndarray,
+    gene_indices: np.ndarray,
+    gene_names: list[str],
+    codeword_gene_names: list[str] | None = None,
+):
+    """Write density maps, gene/codeword categories, and metrics density
+    into an existing zarr group, matching the native Xenium Explorer format.
+
+    Args:
+        root: The zarr root group of transcripts.zarr.zip (already open for writing).
+        xy: Transcript coordinates in microns, shape (n_transcripts, 2).
+        gene_indices: Per-transcript gene index (uint16), shape (n_transcripts,).
+        gene_names: List of gene names.
+        codeword_gene_names: List of codeword names. Defaults to gene_names.
+    """
+    import numcodecs
+    from scipy.sparse import coo_matrix
+
+    if codeword_gene_names is None:
+        codeword_gene_names = gene_names
+
+    n_genes = len(gene_names)
+    n_codewords = len(codeword_gene_names)
+    compressor = numcodecs.Blosc(cname="zstd", clevel=5, shuffle=numcodecs.Blosc.BITSHUFFLE)
+
+    # ── Density grid (10 µm bins) ───────────────────────────────────────
+    grid_size = ExplorerConstants.DENSITY_GRID_SIZE
+    x_min, y_min = xy.min(axis=0)
+    x_max, y_max = xy.max(axis=0)
+
+    origin_x = np.floor(x_min / grid_size) * grid_size
+    origin_y = np.floor(y_min / grid_size) * grid_size
+
+    n_cols = int(np.ceil((x_max - origin_x) / grid_size)) + 1
+    n_rows = int(np.ceil((y_max - origin_y) / grid_size)) + 1
+
+    col_idx = np.floor((xy[:, 0] - origin_x) / grid_size).astype(np.int32).clip(0, n_cols - 1)
+    row_idx = np.floor((xy[:, 1] - origin_y) / grid_size).astype(np.int32).clip(0, n_rows - 1)
+
+    # CSR: rows = gene_index * n_rows + grid_row, cols = grid_col
+    csr_row = gene_indices.astype(np.int64) * n_rows + row_idx.astype(np.int64)
+    density_csr = coo_matrix(
+        (np.ones(len(csr_row), dtype=np.uint16), (csr_row, col_idx)),
+        shape=(n_genes * n_rows, n_cols),
+    ).tocsr()
+    density_csr.data = density_csr.data.astype(np.uint16)
+    density_csr.indices = density_csr.indices.astype(np.uint16)
+    density_csr.indptr = density_csr.indptr.astype(np.uint32)
+
+    density_attrs = {
+        "grid_size": [grid_size, grid_size],
+        "rows": n_rows,
+        "cols": n_cols,
+        "origin": {"x": float(origin_x), "y": float(origin_y)},
+    }
+
+    density_group = root.create_group("density")
+
+    gene_density = density_group.create_group("gene", attributes={**density_attrs, "gene_names": gene_names})
+    gene_density.create_array("data", data=density_csr.data, chunks=(len(density_csr.data),), compressor=compressor)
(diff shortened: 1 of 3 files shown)


sopa/io/explorer/points.py
```diff
@@ -17,6 +17,133 @@ def subsample_indices(indices: np.ndarray, factor: int = 4) -> np.ndarray:
     return np.random.choice(indices, len(indices) // factor, replace=False)
 
 
+def _classify_gene_names(names: list[str], n_cols: int = 9) -> np.ndarray:
```
quentinblampey: Is it possible to simplify this function to avoid the for loop? E.g., using pandas and .apply

sopa/io/explorer/points.py
```diff
@@ -17,6 +17,133 @@ def subsample_indices(indices: np.ndarray, factor: int = 4) -> np.ndarray:
...
+    """
+    import numcodecs
+    from scipy.sparse import coo_matrix
+
+    if codeword_gene_names is None:
+        codeword_gene_names = gene_names
+
+    n_genes = len(gene_names)
+    n_codewords = len(codeword_gene_names)
+    compressor = numcodecs.Blosc(cname="zstd", clevel=5, shuffle=numcodecs.Blosc.BITSHUFFLE)
```
quentinblampey: Is there a specific reason to use a compressor here instead of the default (`auto`) compressor?

sopa/io/explorer/points.py
```diff
@@ -17,6 +17,133 @@ def subsample_indices(indices: np.ndarray, factor: int = 4) -> np.ndarray:
...
+            cats[i, 0] = True
+    return cats
+
+
+def _write_density(
+    root: zarr.Group,
+    xy: np.ndarray,
+    gene_indices: np.ndarray,
+    gene_names: list[str],
+    codeword_gene_names: list[str] | None = None,
```
quentinblampey: `_write_density` is always called with `codeword_gene_names = None`, so this argument can either be removed, or passed as an argument at least once

sopa/io/explorer/points.py
```diff
@@ -17,6 +17,133 @@ def subsample_indices(indices: np.ndarray, factor: int = 4) -> np.ndarray:
...
+
+    # ── Metrics density (20 µm bins) ────────────────────────────────────
+    spacing = ExplorerConstants.METRICS_DENSITY_SPACING
+    mx_count = int(np.ceil((x_max - origin_x) / spacing)) + 1
+    my_count = int(np.ceil((y_max - origin_y) / spacing)) + 1
+
+    m_col = np.floor((xy[:, 0] - origin_x) / spacing).astype(np.int32).clip(0, mx_count - 1)
+    m_row = np.floor((xy[:, 1] - origin_y) / spacing).astype(np.int32).clip(0, my_count - 1)
+
+    metrics = np.zeros((my_count, mx_count, 4), dtype=np.float32)
```
quentinblampey: `metrics[:, :, 3]` always contains only 0, is it expected?

sopa/io/explorer/points.py
```diff
@@ -17,6 +17,133 @@ def subsample_indices(indices: np.ndarray, factor: int = 4) -> np.ndarray:
...
+    }
+
+    density_group = root.create_group("density")
+
+    gene_density = density_group.create_group("gene", attributes={**density_attrs, "gene_names": gene_names})
+    gene_density.create_array("data", data=density_csr.data, chunks=(len(density_csr.data),), compressor=compressor)
+    gene_density.create_array("indices", data=density_csr.indices, chunks=(len(density_csr.indices),), compressor=compressor)
+    gene_density.create_array("indptr", data=density_csr.indptr, chunks=(len(density_csr.indptr),), compressor=compressor)
+
+    cw_density = density_group.create_group("codeword", attributes={**density_attrs, "codeword_names": codeword_gene_names})
```
quentinblampey: `cw_density` here is just a duplicate of `gene_density`, right?

quentinblampey (changes requested):
Hi @jackytamkc,

Thanks for the PR, it looks great! I made some comments, but it's relatively minor.

By any chance, have you already tested it on large datasets? If not, I may run a quick benchmark to see if it scales well, but the code suggests it will

sopa/io/explorer/points.py
jackytamkc: Yes. replaced with a pandas Series.str.contains approach.

sopa/io/explorer/points.py
jackytamkc: The reason behind it was reverse-engineered from native Xenium zarrs but realizing zarr default also works ( and consistent with the rest of the code), can safely use the default.

sopa/io/explorer/points.py
jackytamkc: Can remove.

sopa/io/explorer/points.py
jackytamkc: Expected. channel 3 is 'is_negative_control_codeword' which has no source signal at the writer level under unified spatialdata schema.

sopa/io/explorer/points.py
jackytamkc: Yes. It was there also because of reverse-engineering using XOA 4.0 reference which has the cw_density but XOA3.0 also works fine (which has no cw_density). Tested and it's still work.

jackytamkc:
Hi @quentinblampey,

Thanks. I haven't tried on extremely large dataset (this was tested on CosMx data at 55M transcripts). Worth to try but I assume it should scale pretty well.

Cheers,
Jacky

fentouxungui:
@quentinblampey 

Thank you for your prompt reply.

The points I mentioned earlier and the code I used were both generated by Claude Code. I apologize that I am unable to verify the accuracy of the code, as I am not proficient in Python programming. I will now walk you through the issues I encountered step by step, starting from the beginning. 

First, I will use a public dataset from the 10×Genomics website to illustrate the problem I discovered.
Xenium Prime 5K Data (total 200k cells): https://www.10xgenomics.com/datasets/xenium-comparison-fresh-frozen-human-ovarian-cancer. Please select “Xenium Prime 5k sample” under “Per sample output” on the right, and download the Xenium output bundle.

After downloading, open the data with XeniumExplorer. When you query the expression of a specific gene, the software runs very smoothly—you won’t notice any lag even when scrolling while transcripts are loading. 

Then select a region (e.g., approximately 20k cells) and export a GeoJSON file. Use the SOPA software to extract the data, and open the extracted results in XeniumExplorer. It correctly identifies all SOPA output files, but when you query the transcript expression of a specific gene, there is noticeable lag. For example, with the EPCAM gene, when you zoom in on the image and scroll left or right, there is a noticeable delay, though the software does not freeze. 

My extracted data and codes used can be found in the “demo-data-1” folder on Google Drive (https://drive.google.com/drive/folders/1KrVH75ldBlx8ptQNX25ERKnQP26pAYZN?usp=drive_link).

I believe XeniumExplorer would perform even worse when extracting a subset from a large dataset. For example, in my own **Xenium Prime 5k** data, my goal was to extract 160k cells from 700k cells. When I opened the extracted data in XeniumExplorer and queried gene expression, the software froze almost immediately. Therefore, I believe that the transcript output file `transcripts.zarr.zip` generated by SOPA has certain performance issues.


Currently, the two scripts generated by ClaudeCode—regenerate-transcripts.py and regenerate-transcripts-with-density.py ([PR411](https://github.com/gustaveroussy/sopa/pull/411), thanks to [jackytamkc](https://github.com/jackytamkc))—have successfully helped me obtain a transcripts.zarr.zip file that is fully compatible with XeniumExplorer. The code can be found in the **successful-extraction-codes-by-ClaudeCode** directory. Since I am not proficient in Python programming, I am unable to answer each of your questions; I apologize for this.



Translated with DeepL.com (free version)

quentinblampey (approved):
Thanks @jackytamkc, I'll merge!

PR #416, opened 2026-04-20
[feature][adding visium hd in snakemake]

reJELIN:
First pull request ever please be kind:

[Visium_hd](https://github.com/prism-oncology/sopa/issues/415) is not natively taken in the snakemake pipeline, here I propose a brief fix. Your data_path, should be the spaceranger output.

I've change the shell of the to_spatialdata. This rule introduces a "fork in the road." By using a conditional statement (if/else), the rule inspects the technology variable at runtime. It decides whether to use the standard sopa command or a custom Python script for visium hd.

The shell block becomes a clean placeholder: `{params.shell_command}`. This keeps the execution part of the rule very tidy.

Unfortunately it's not perfect because we're loosing the CLI for something more hybrid but it is still a good workaround.

```python3
rule to_spatialdata:
    input:
        [] if config["is_toy_reader"] else paths.data_path,
    output:
        paths.sdata_json_attrs if paths.data_path else [],
    conda:
        "sopa"
    params:
        shell_command = lambda wildcards, input :(f"sopa convert {paths.sdata_path} --sdata-path {paths.sdata_path,} {args['read'].as_cli()}" if technology !="visium_hd" else f"python3 {src_dir}/to_spatialdata_visium_hd.py --input_path {paths.sdata_path} --output_path {paths.sdata_path} --fullres_image_file {highres_image}")
    shell:
        """
        {params.shell_command}
        """
```

As you can see it add the necessity to add 3 new variables:

```python3
technology = config["read"]["technology"]
src_dir = os.path.join(workflow.basedir,"src")
highres_image = config.get("highres_image", None)
```
`highres_image` is your path to your high resolution image, by default with it is none.

i'm not sure why `args['read'].as_cli()` exists ? isn't it simple to just use config["read"]["technology"] ? in anycase I made an if/else statement base on the technology name.

`src_dir` is sample it is the path to scr directory.

It is not perfect and it would probably be better to integrate this directly into sopa convert command line but at least it work.

### DIFF
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -4,6 +4,10 @@ config = validate_config(config) # validate the Snakemake config
 paths = WorkflowPaths(config) # object handling the paths to the files that will be created
 args = Args(paths, config) # object handling the arguments passed to the workflow
 
+technology = config["read"]["technology"]
+src_dir = os.path.join(workflow.basedir,"src")
+highres_image = config.get("highres_image", None)
+
 ### Segmentation rules
 include: "rules/utils.smk"
 include: "rules/cellpose.smk"
@@ -42,12 +46,10 @@ rule to_spatialdata:
     conda:
         "sopa"
     params:
-        reader = args['read'].as_cli(),
-        data_path = paths.data_path,
-        sdata_path = paths.sdata_path,
+        shell_command = lambda wildcards, input :(f"sopa convert {paths.sdata_path} --sdata-path {paths.sdata_path,} {args['read'].as_cli()}" if technology !="visium_hd" else f"python3 {src_dir}/to_spatialdata_visium_hd.py --input_path {paths.sdata_path} --output_path {paths.sdata_path} --fullres_image_file {highres_image}")
     shell:
         """
-        sopa convert {params.data_path} --sdata-path {params.sdata_path} {params.reader}
+        {params.shell_command}
         """
 
 rule tissue_segmentation:
--- a/workflow/src/to_spatatial_visium_hd.py
+++ b/workflow/src/to_spatatial_visium_hd.py
@@ -1,0 +1,24 @@
+import argparse
+import sopa.io
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument("--input_path",
+                     type = str,
+                     help = 'path to the output of Space ranger',
+                     required = True)
+
+parser.add_argument("--output_path",
+                     type = str,
+                     help = 'path to the output of sopa to save .explorer and .zarr folders',
+                     required = False)
+
+parser.add_argument("--fullres_image",
+                     type = str,
+                     help = 'path to the fullres image file',
+                     required = False)
+
+
+sdata = sopa.io.visium_hd(path = args.input_path, fullres_image_file = args.fullres_image,dataset_id = "")
+
+sdata.write(f"{args.output_path}", overwrite=True)


quentinblampey:
Hi @reJELIN,

Thanks for the PR!

Actually, `config["read"]` may contain other attributes than `"technology"`, see for instance [this config](https://github.com/prism-oncology/sopa/blob/main/workflow/config/visium_hd/proseg_10x_website.yaml). Here, `.as_cli()` will convert the whole dictionary structure into a string that can be used by the Sopa CLI.

Based on this, I think you can use `.as_cli()` to pass the `fullres_image_file` directly, without having to create an extra python script. I think that all you need to do is store the `fullres_image_file` here:

```python
config["read"]["kwargs"]["fullres_image_file"] = fullres_image_file
```
And then, the `.as_cli()` should do the job!

Actually, I'm wondering if we can't even just pass `fullres_image_file` directly to the snakemake command? Something like this (I didn't check the syntax is right): 
```sh
--config read:kwargs:fullres_image_file="image.tif"
```

reJELIN:
If I understood correctly, the `convert` command line is using the SpatialData object to write the file into the correct format using SpatialData.Write function ? 

the important line is in the `convert` function:

```python3
sdata: SpatialData = getattr(io, technology)(data_path, **kwargs) 
```

it is then calling the function `visium_hd` in `sopa.io.reader` ? 

won't this work  `sdata: SpatialData = getattr(io, technology)(data_path,fullres_image_file, **kwargs)` if we're adding an optional argument to give the path to the H&E image to the `convert` function ?

quentinblampey:
Yes, `getattr(io, technology)(data_path, **kwargs)` gets the right reader based on its name. E.g., if `technology = "visium_hd"`, then it calls `sopa.io.visium_hd(data_path, **kwargs)`

This is why I believe passing the `fullres_image_file` through the kwargs should work:
doing `config["read"]["kwargs"]["fullres_image_file"] = fullres_image_file`
will then call `sopa.io.visium_hd(data_path, fullres_image_file=fullres_image_file)` .

> won't this work sdata: SpatialData = getattr(io, technology)(data_path,fullres_image_file, **kwargs) if we're adding an optional argument to give the path to the H&E image to the convert function ?

It would, yes, but then we add an extra argument to the CLI, and it adds extra logic to ensure backward compatibility (we want to make it still work for users passing `fullres_image_file` through the kwargs, if any).

In any case, I'll have to clarify this use case has in the Sopa snakemake docs :)

reJELIN:
I'm going to test adding `--config read:kwargs:fullres_image_file="image.tif"` to the snakemake command.

reJELIN:
Indeed it work here is the correct way to add the fullres_image_file:

```bash
snakemake -s /<path_to_sopa>/workflow/Snakefile     \
--config     data_path="<mypath_to_spaceranger_results>/outs"     read='{"kwargs": {"fullres_image_file": "<mypath_to_the_high_res_image>.ome.tif"}}'    \
--configfile /<path_to_sopa>/config/visium_hd/stardist.yaml     \
--workflow-profile /<path_to_sopa>/2.2.5/workflow/profile/slurm-igr 
```

I've notice a quick "bug":

NameError in file /sopa/2.2.5/workflow/rules/proseg.smk, line 3:
name 'self' is not defined
  File "/sopa/2.2.5/workflow/rules/proseg.smk", line 3, in <module>

Is it specific to using the visium_hd ? I comment it and it work.

quentinblampey:
Indeed, it was a bug, probably due to the recent refactoring I made in the snakemake pipeline
Thanks for catching it, I fixed it, could you please `git pull`?

Alright, good to know that `read='{"kwargs": {"fullres_image_file": "<mypath_to_the_high_res_image>.ome.tif"}}'` works!
The syntax is not really convenient though, maybe we could try the option I proposed above, i.e. adding a `fullres_image_file` param in the config, and add it at the right place in the config validation. It could be something like that:

```python
if "fullres_image_file" in config:
    config["read"]["kwargs"] = config["read"].get("kwargs", {})
    config["read"]["kwargs"]["fullres_image_file"] = config["fullres_image_file"]
```

What do you think? This way we could directly pass `fullres_image_file=...` to the snakemake command

reJELIN:
Yep your fix work, thank you for the correction !

Indeed this solution should definitely works

reJELIN:
Indeed adding this solution works well:

```python3
from src import WorkflowPaths, Args, validate_config

config = validate_config(config) # validate the Snakemake config
paths = WorkflowPaths(config) # object handling the paths to the files that will be created
args = Args(paths, config) # object handling the arguments passed to the workflow

if "fullres_image_file" in config:
    config["read"]["kwargs"] = config["read"].get("kwargs", {})
    config["read"]["kwargs"]["fullres_image_file"] = config["fullres_image_file"]

```
Thank you very much for your help !

quentinblampey:
Great! Do you want to make a PR or can I just update it myself directly?

EDIT: I can do it, and I'll also mention it inside the docs!

reJELIN:
You found the solution, so it would be logical that you do the PR.

PR #418, opened 2026-05-08
feat(segmentation): add scale argument to run cellpose/stardist on coarser multi-scale levels

jackytamkc:
Closes #XXX 

## What this changes

Adds an optional `scale: str = "scale0"` argument that threads through `get_spatial_image` → `Patches2D` → `make_image_patches` → `StainingSegmentation` → `custom_staining_based` → `cellpose`. Default is `"scale0"` everywhere, so existing call sites are unchanged.

## Motivation

Modern transcript-resolution platforms ship images at sub-cellular pixel sizes that exceed what segmenters need or what GPUs can handle:

| Platform | Native pixel size |
|---|---|
| Xenium (DAPI) | ~0.2125 µm/px |
| CosMx | ~0.12028 µm/px |

On a CosMx liver biopsy I segmented (28074 × 96205 px = 2.7 GP at scale0), Cellpose-SAM v4 OOMs on a 95 GB RTX 6000 Pro Blackwell. The forward pass tiles cleanly but `cellpose.dynamics.masks_to_flows_gpu` allocates a `(2, 9, N) int32` neighbours tensor where `N` is cell-pixel count; at ~30% cell coverage that one tensor is ~58 GB, plus the int64 padded mask (~21 GB) and the model (~5 GB) — over budget.

The same image at scale1 (14037 × 48102 = 0.67 GP) finishes cleanly with peak GPU ~50 GB and produces ~33,000 cells. Quality is indistinguishable because cellpose's internal rescale was discarding the scale0 detail anyway.

`make_image_patches(patch_width=…)` already supports tiling scale0, but for users with one big patch and a single GPU, running directly on a coarser scale is cleaner and avoids the per-patch model-load overhead in current sopa.

## API

```python
sopa.make_image_patches(sdata, image_key="segmentation",
                        patch_width=None, scale="scale1")
sopa.segmentation.cellpose(sdata, channels=[...], image_key="segmentation",
                           pretrained_model=..., diameter=60, min_area=400,
                           scale="scale1")
```

`diameter`, `patch_width`, `min_area` are interpreted in **pixels of the chosen scale** (the same convention sopa already uses for scale0). At `scale1` (×2 downsample) divide your scale0 values by 2 (linear) or 4 (area).

Output cell boundaries inherit the chosen scale's `intrinsic → global` transformation, so they map back to the global frame for downstream transcript assignment without any user-side rescale.

## Files changed

- `sopa/utils/utils.py` — `_return_element`, `get_spatial_element`, `get_spatial_image` accept `scale`
- `sopa/patches/patches.py` — `Patches2D.__init__` accepts `scale`, no longer hardcodes `"scale0"`
- `sopa/patches/factory.py` — `make_image_patches` accepts `scale`, threads to `Patches2D`
- `sopa/segmentation/stainings.py` — `StainingSegmentation.__init__` and `add_shapes` accept `scale`; spatial slicing in `_run_patch` switched to `.isel()` (see gotcha below)
- `sopa/segmentation/methods/_custom.py` — `custom_staining_based` accepts `scale`
- `sopa/segmentation/methods/_cellpose.py` — `cellpose()` accepts `scale`

## Important gotcha — `.sel` vs `.isel` in `_run_patch`

In the process of testing this I hit a real bug worth a separate look even outside this PR:

`spatialdata` multi-scale images store coord values **in scale0-pixel-equivalent units** at every level — for `segmentation/scale1["image"].coords["x"].values` returns `[1.0, 3.0, 5.0, …, 96204.0]`, not `[0, 1, 2, …, 48101]`.

`Patches2D._init_patches` builds patch bounds with `xmax = len(coords["x"])` (pixel-index space). `StainingSegmentation._run_patch` previously sliced the image with:

\`\`\`python
image = self.image.sel(c=self.channels,
                        x=slice(bounds[0], bounds[2]),
                        y=slice(bounds[1], bounds[3])).values
\`\`\`

xarray's `.sel(x=slice(a, b))` is a coord-value range query, not an index range. At scale0 these accidentally agree (coord values 0.5–96204.5 fall within `[0, 96205]`). At scale1 they don't — `.sel(x=slice(0, 48102))` selects only ~half the pixels (the columns whose coord values are < 48102).

This PR switches the spatial slicing to `.isel()` (integer index), keeping `.sel(c=…)` for channel-name selection:

\`\`\`python
image = self.image.sel(c=self.channels).isel(
    x=slice(bounds[0], bounds[2]),
    y=slice(bounds[1], bounds[3]),
).values
\`\`\`

That makes `Patches2D`'s pixel-index bounds consistent with the slicing in `_run_patch` regardless of what coord values the underlying DataArray carries. **No behavioural change at scale0; it just makes scale ≠ 0 work.**

I checked the other call sites in sopa for the same `.sel(x=…, y=…)` pattern on multi-scale images; none others surfaced. Worth flagging for review since it's subtle and could have been missed silently in any current scale > 0 use case.

## Verification

Tested on a CosMx liver biopsy (`stitched_image` 28074 × 96205 px scale0):

- **scale0 + flow_threshold=2**: OOMs at 187 s on a 95 GB GPU (matches what motivated this work).
- **scale1 + flow_threshold=2**: completes in 198 s with peak GPU 49.55 GB; produces ~33,000 cells whose centroids span the full tissue extent (`x ∈ [3969, 44300]` in scale1-px, equivalent to the full tissue region after applying the scale1 → global transform).
- **scale1 cell distribution matches segmentation-channel intensity** in a 5×5 tile heatmap — i.e. cells appear where there is tissue, not where there is background, so the segmentation is biologically sensible at the coarser scale.

Backward compatibility: all defaults are `"scale0"`, so existing tests and tutorials are unaffected.

## Open questions for review

- Parameter name: I went with `scale` since it matches the spatialdata convention. `scale_level` or `scale_key` would also read fine — happy to rename if either is preferred.
- `stardist` and `baysor` callers: I have not touched those in this PR (`baysor` is transcript-based so the parameter does not apply; `stardist` calls `custom_staining_based` so it gets the support automatically through the existing inheritance, but the public `stardist()` function would need an explicit kwarg pass-through to surface it). Want me to add that to this PR or leave it for a follow-up?
- I added the `.sel` → `.isel` fix in this same PR because anything coarser than scale0 is broken without it. Happy to split into two PRs if you prefer the bug fix to land independently.

## Checklist

- [ ] Tests pass locally (ran segmentation tests; not the full suite)
- [ ] Documented the new parameter in docstrings of all six entry points
- [x] Backward-compatible (default `"scale0"`)
- [x] No new dependencies
- [x] Tested on real data (CosMx biopsy, scale0 OOMs vs scale1 succeeds)

### DIFF
--- a/sopa/patches/factory.py
+++ b/sopa/patches/factory.py
@@ -18,18 +18,20 @@ def make_image_patches(
     image_key: str | None = None,
     roi_key: str | None = SopaKeys.ROI,
     key_added: str | None = None,
+    scale: str = "scale0",
 ):
     """Create overlapping patches on an image. This can be used for image-based segmentation methods such as Cellpose, which will run on each patch.
 
     Args:
         sdata: A `SpatialData` object.
-        patch_width: Width of the patches, in pixels. If `None`, creates only one patch.
+        patch_width: Width of the patches, in pixels of the chosen `scale`. If `None`, creates only one patch.
         patch_overlap: Number of pixels of overlap between patches.
         image_key: Optional key of the image on which the patches will be made. If not provided, it is found automatically.
         roi_key: Optional name of the shapes that need to touch the patches. Patches that do not touch any shape will be ignored during segmentation. By default, uses `"region_of_interest"` if existing. If `None`, all patches will be used.
         key_added: Optional name of the patches to be saved. By default, uses `"image_patches"`.
+        scale: For a multi-scale image, scale level to use (e.g. `"scale0"`, `"scale1"`). Patches are created in the chosen scale's intrinsic pixel space; their stored transformation maps back to the global coordinate system. When you later run a staining-based segmentation, pass the same `scale` to that call so the image is sliced at matching coords.
     """
-    image_key, _ = get_spatial_image(sdata, key=image_key, return_key=True)
+    image_key, _ = get_spatial_image(sdata, key=image_key, return_key=True, scale=scale)
 
     patches = Patches2D(
         sdata,
@@ -37,6 +39,7 @@ def make_image_patches(
         patch_width=patch_width,
         patch_overlap=patch_overlap,
         roi_key=roi_key,
+        scale=scale,
     )
 
     patches.add_shapes(key_added=key_added)
--- a/sopa/patches/patches.py
+++ b/sopa/patches/patches.py
@@ -81,6 +81,7 @@ class Patches2D:
         patch_overlap: float | int = 50,
         roi_key: str | None = SopaKeys.ROI,
         use_roi_centroids: bool = False,
+        scale: str = "scale0",
     ):
         """
         Args:
@@ -90,6 +91,7 @@ class Patches2D:
             patch_overlap: Overlap width between the patches
             roi_key: Optional name of the shapes that need to touch the patches. Patches that do not touch any shape will be ignored. If `None`, all patches will be used.
             use_roi_centroids: If `True`, the ROI will be computed from the centroids of the shapes in `roi_key`. If `False`, the ROI will be computed from the shapes themselves.
+            scale: For a multi-scale image, scale level to use for patch creation (e.g. `"scale0"`, `"scale1"`). Patch bounds and the stored transformation will follow this scale's intrinsic coordinate system.
         """
         patch_width = float("inf") if (patch_width is None or patch_width == -1) else patch_width
 
@@ -98,9 +100,15 @@ class Patches2D:
         self.sdata = sdata
         self.element = sdata[element] if isinstance(element, str) else element
         self.original_element = self.element  # in case the element is a DataTree
+        self.scale = scale
 
         if isinstance(self.element, DataTree):
-            self.element = next(iter(self.element["scale0"].values()))
+            if scale not in self.element:
+                available = list(self.element.keys())
+                raise ValueError(
+                    f"Scale '{scale}' not found in image. Available scales: {available}"
+                )
+            self.element = next(iter(self.element[scale].values()))
 
         if isinstance(self.element, DataArray):
             xmin, ymin = 0, 0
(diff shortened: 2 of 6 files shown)


quentinblampey:
Hi @jackytamkc,

Thanks for opening, this is relevant :)
Is there a reason why you closed the Pull Request?

Do you know any Cellpose resources that explain that scale0 may be overkill, and which scale is recommended? Or is it a personal observation?

We can work on this - FYI @alihamraoui

jackytamkc:
Hi Quentin,

I was originally planning to open just an issue (and my ai accidentally push this PR under my radar :/) to see if you are interested to add this feature given people are more willing to use newer cpsam rather than cp3. I didn't check cellpose resources but it's based on my personal observation on in-house cosmx data, where it is exceptionally high resolution that most segmenting tools do not need. And transcript spillover in spatial data is common so people (including me) tend to correct them using a transcript-based method afterwards so the upstream imaging step is not necessarily needed to be super accurate. 

On a WSI image segmenting at scale0 (28074 × 96205px I have tested, on a blackwell 96GB GPU) will definitely OOM on GPU. While I fully appreciate the resolve conflict step some people may still prefer segmenting on a whole tissue level if GPU permits. If you think this is relevant I can reopen the PR :)

Cheers,
Jacky

quentinblampey:
Okay, thanks for your insight! I added it to our task board, we'll work on this (it will likely take a few weeks though, because we have a few high-priority tasks currently)

Just out of curiosity, since you run a transcript-based method afterward, why not use the default CosMx segmentation as a prior instead of running cellpose?

jackytamkc:
That's totally fine! 

The default cosmx segmentation has a border stitching effect, since they image FOV by FOV, and cells between FOV will be inevitably splitted into 2. I tend to use proseg afterwards and proseg depends on prior heavily and does not really merge or introduce new cells, which means that stitching effect will stay on in the final output.

Also in general I think cpsam segmentation gives better and more accurate output than default cosmx segmentation. Not sure about xenium since I don't have xenium data yet.

quentinblampey:
Totally makes sense, thanks again for your insight!😊

quentinblampey:
Re-opening since I'll work on this soon
EDIT: actually, I can't, the branch has been deleted. I'll make a new branch

jackytamkc:
Thanks. Let me know if you need help with anything!

PR #425, opened 2026-05-29
Increase point write speed for sopa.io.explorer.write

pakiessling:
Hi Quentin,

I started using Sopa again after quite some time and I noticed that writing transcripts for the explorer was way slower than I remembered. For an average dataset of mine it takes > 12 hours now.

After some investigation this seems to be connected to the update to Zarr v3 in 96832a1 and 1fe1bee.

This slowed transcript writing as it added an additional loop over the data to build GRIDS_ATTRS.
In addition zarr 3 create_array seems to have quite a lot more overhead than zarr 2.

In this draft I made some changes that increase speed by a lot.

speed_comparison_v2

Summary what was changed:

- Instead of a separate loop over the tiles to build GRIDS_ATTRS we can have a single write loop and then append grids/.zattrs at the end
- `np.where(tiles_str_indices == str_index)` was scanning all transcripts per tile, which can be avoided with a groupby
- As the structure of transcripts.zarr.zip is very simple it can be generated without the high level API which would call a lot of unccesary `_open_to_write()` on the Zip

Disclaimer: This was LLM assisted as I don't know much about .zarr internals

I verified that this change does not affect the output transcripts.zarr.zip, It can still be read by Explorer without any issues.

### DIFF
--- a/sopa/io/explorer/points.py
+++ b/sopa/io/explorer/points.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from math import ceil
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


quentinblampey:
Hi @pakiessling,

Thanks for reporting - this is a known issue: moving to zarr v3 was complicated, and this was the only working solution I came up with. But I was already planning on trying to avoid this loop, so thanks for taking the lead and working on this before me!!

I'll have a look at the PR soon (hopefully in the next two weeks) 👍

sopa/io/explorer/points.py
```diff
@@ -1,17 +1,25 @@
...
 import zarr
+from numcodecs import Blosc
 from zarr.storage import ZipStore
 
 from ._constants import ExplorerConstants, FileNames
 from .utils import explorer_file_path
 
 log = logging.getLogger(__name__)
 
+_BLOSC = Blosc(cname="lz4", clevel=5, shuffle=Blosc.SHUFFLE)
```
quentinblampey: I checked a few Xenium files manually, and I always got `zstd` for the compressor. Are you sure the compression was `lz4` for the files you checked?

It looks like it's working though, so I don't really know if it matters

quentinblampey (commented):
Hi @pakiessling,

Thanks again for the PR, it's amazing! The code logic for extracting the different levels remains the same, but only the writing logic is updated, and it seems to work.

I pushed a minor cleanup, and I also left one comment (see below):

sopa/io/explorer/points.py
pakiessling: I checked a few more of my samples. It looks like they changed from lz4 to zstd at some point, lets go with zstd

pakiessling:
Sweet,
should we add some tests that write output and try to read it back in?

quentinblampey:
Yes I'll add some tests!

Regarding `zstd`: by any chance, would you be able to easily add the `zstd` option to your benchmark @pakiessling? To see if it's much slower than `lz4` or not

pakiessling:
speed_comparison_v2

Looks like zstd is ~30–50% slower but archieves ~10–30% smaller size on disk

quentinblampey:
Great, thanks! In that case, I think we'll go with zstd: it's not too slow, and it's consistent with the new Xenium files  + creates smaller files 

I'll add some tests and merge the PR soon, likely this week

quentinblampey (approved):

PR #426, opened 2026-05-31
Add `nucleus_shapes_key` to the writer

### LINKED ISSUE #182: [Feature] Allow two separate categories of shapes to be written to Xenium Explorer
The default Xenium output included sergmented nuclei and cells that can be observed in the Explorer.

With sopa you are able to segment arbitrary shapes: cells, nuclei, organelles.
However it is currently not possible to add more than one type of shape to the Xenium explorer with Sopa.

It would be cool to allow the user to write arbitrary shapes to the Xenium explorer "nuclei slot"

quentinblampey:
Previously, `explorer.write` was writing only cell boundaries, not the nucleus. This PR adds an optional `nucleus_shapes_key` to denote the name of the GeoDataFrame containing the cell nuclei.
Note that both GeoDataFrames' indices should match so that we can map every cell to a nucleus.

FYI @pakiessling and @lzyywzq, happy to get some feedback if you have some time to try it out using this branch name (`nucleus_shapes_key`)

Closes #182


sopa/io/explorer/converter.py
```diff
@@ -144,7 +146,15 @@ def write(
         if preserve_ids is None:
             preserve_ids = _update_preserve_ids(geo_df.index)
 
-        write_polygons(path, geo_df, polygon_max_vertices, pixel_size=pixel_size, preserve_ids=preserve_ids)
+        geo_df_nucleus = None
+        if nucleus_shapes_key is not None:
```
alihamraoui, line 150: Mybe add a verification for nucleus_shapes_key? like:
```python
if nucleus_shapes_key is not None:
    assert nucleus_shapes_key in sdata.shapes, (
        f"No '{nucleus_shapes_key}' found in `sdata.shapes`. "
        f"Available shapes: {list(sdata.shapes.keys())}"
    )
```

alihamraoui (approved):

PR #430, opened 2026-06-24
Support the Baysor Cpp implementation

### LINKED ISSUE #427: [Bug] Baysor segmentation fails: missing config.toml
## Description
I am running Sopa 2.2.8 on macOS and following the API tutorial for Baysor segmentation:

sopa.segmentation.baysor(sdata, min_area=10)

Sopa reaches the Baysor call but fails with:

CalledProcessError:
Command '/Users/.../.julia/bin/baysor run --polygon-format GeometryCollection -c config.toml transcripts.csv :sopa_prior'
returned non-zero exit status 0

## Environment
OS: macOS
Sopa version: 2.2.8
Baysor: custom install (C++ build, run subcommand available)
Python environment: conda
Julia: 1.11 installed (via juliaup)

alihamraoui:
This PR adds support for Baysor C++ v0.8.2 (issue #427).

### What changed

- Added `-o .` to the Baysor command so the outputs are written directly into the patch directory instead of a `segmentation.csv/ `subfolder.

### Reverted changes

- Re-read `segmentation_counts.loom `with loompy using `validate=False` and rebuild the `AnnData` with the correct orientation. The Baysor C++ implementation writes the matrix transposed (cells × genes) relative to the Loom spec (genes × cells), which makes `read_loom` reject the file.

Loom is not well maintained and supports only a subset of anndata features. The `anndata.io.read_loom `reader is also deprecated since anndata 0.13 (see https://anndata.readthedocs.io/en/latest/generated/anndata.io.read_loom.html). Switching the Baysor command to TSV output might be the better long-term option.

@quentinblampey
 
Closes #427

### DIFF
--- a/sopa/segmentation/methods/_baysor.py
+++ b/sopa/segmentation/methods/_baysor.py
@@ -146,7 +146,7 @@ def _get_baysor_command(prior_shapes_key: str | None) -> str:
 
     prior_suffix = f" :{prior_shapes_key}" if prior_shapes_key else ""  # use a prior segmentation
 
-    return f"{baysor_executable_path} run {polygon_format} -c config.toml transcripts.csv" + prior_suffix
+    return f"{baysor_executable_path} run {polygon_format} -c config.toml transcripts.csv -o ." + prior_suffix
 
 
 def _use_polygons_format_argument(baysor_executable_path: str) -> bool:
--- a/sopa/segmentation/methods/utils.py
+++ b/sopa/segmentation/methods/utils.py
@@ -87,6 +87,38 @@ def resolve(
 
     log.info(f"Added sdata.tables['{SopaKeys.TABLE}'], and {len(geo_df)} cell boundaries to sdata['{key_added}']")
 
+def _read_baysor_loom(loom_file: Path, obs_names: str, var_names: str) -> AnnData:
+    """Read a Baysor loom into an `AnnData`.
+
+    looms written with a transposed `/matrix` (cells x genes instead of 
+    genes x cells, as the Baysor Cpp does) can still be read.
+    """
+    import loompy
+
+    with loompy.connect(str(loom_file), "r", validate=False) as ds:
+        matrix = ds[:, :]
+        gene_names = np.asarray(ds.ra[var_names]).astype(str)
+        col_attrs = {key: np.asarray(ds.ca[key]) for key in ds.ca.keys()}
+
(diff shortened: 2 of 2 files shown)


sopa/segmentation/methods/_baysor.py
```diff
@@ -146,7 +146,7 @@ def _get_baysor_command(prior_shapes_key: str | None) -> str:
 
     prior_suffix = f" :{prior_shapes_key}" if prior_shapes_key else ""  # use a prior segmentation
 
-    return f"{baysor_executable_path} run {polygon_format} -c config.toml transcripts.csv" + prior_suffix
+    return f"{baysor_executable_path} run {polygon_format} -c config.toml transcripts.csv -o ." + prior_suffix
```
quentinblampey, line 149: Does it also work with the previous baysor versions? In baysor<0.8, `-o` is the path to one single file, so I'm not sure what happens if the output file is `.`?

quentinblampey (commented):
Two naive questions (because I didn't try the new version):
1. The original issue mentions `Baysor segmentation fails: missing config.toml`, but in the command line you only change the output location, so how did it fix the config.toml issue? Or maybe the issue the user had was not related to the new baysor version?
2. One other question about backward compatibility (see my review comment)

> Switching the Baysor command to TSV output might be the better long-term option.

If Baysor always output a TSV (in both >=0.8 and <0.8 versions) then, yes definitely, it would avoid this read_loom complexity

sopa/segmentation/methods/_baysor.py
alihamraoui: Yes, I tested it on `v0.7.1` and `cpp_v0.8.2`, and it works. The `-o ` option specifies the output path. In `cpp_v0.8.2`, however, it defaults to `segmentation.csv`, which is a bit confusing. 😕

alihamraoui:
I'm not sure why the user mentioned a missing config.toml.

I think the error message `non-zero exit status 0 ` actually indicates that the command executed successfully, but the output was not what `Sopa` expected.

I was able to reproduce the same message simply by switching to the Baysor C++ version:

```
Command '/Users/.../.julia/bin/baysor run --polygon-format GeometryCollection -c config.toml transcripts.csv :sopa_prior'
returned non-zero exit status 0
```

The fix resolves the issue. But let's switch to `TSV` output by adding` --count-matrix-format tsv` to the Baysor command. That should work for all versions.

quentinblampey (approved):

PR #432, opened 2026-07-06
Enable using a scale>0 in staining-based segmentation for CLI/pipeline

### LINKED ISSUE #424: [Feature] Enable `scale` arg in CLI/pipeline for staining-based segmentation
Follow-up issue - see [this PR for extra context](https://github.com/prism-oncology/sopa/pull/423) (do the same for the CLI and the pipelines)

alihamraoui:
Enable using a scale > 0 in staining-based segmentation for CLI/pipeline

Adds a --scale option to sopa patchify image so staining-based segmentation (e.g. Cellpose) can run on a lower-resolution scale instead of always using the highest one. See #423.

```bash
sopa patchify image tuto.zarr --scale "scale1"  ...
```

This is exposed through the Snakemake pipeline via a new scale key in the patchify config:

```python
patchify:
  scale: "scale"
```

Closes https://github.com/prism-oncology/sopa/issues/424

### DIFF
--- a/workflow/config/merscope/baysor_cellpose.yaml
+++ b/workflow/config/merscope/baysor_cellpose.yaml
@@ -3,6 +3,7 @@ read:
   technology: merscope
 
 patchify:
+  scale: "scale0"
   patch_width_pixel: 6000
   patch_overlap_pixel: 150
   patch_width_microns: 1000
--- a/workflow/config/merscope/baysor_cellpose.yaml
+++ b/workflow/config/merscope/baysor_cellpose.yaml
@@ -3,7 +3,6 @@ read:
   technology: merscope
 
 patchify:
-  scale: "scale0"
   patch_width_pixel: 6000
   patch_overlap_pixel: 150
   patch_width_microns: 1000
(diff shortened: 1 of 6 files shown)


workflow/config/merscope/baysor_cellpose.yaml
```diff
@@ -3,6 +3,7 @@ read:
   technology: merscope
 
 patchify:
+  scale: "scale0"
```
quentinblampey: I think it's not needed to add this, since it's already the default value
Or you wanted to add this for clarity for the user?

quentinblampey (commented):
Looking good @alihamraoui, thanks!

workflow/config/merscope/baysor_cellpose.yaml
alihamraoui: I added it for clarity, but the workflow will break if a single-scale image is provided, as in the toy data, so I'll delete it.
Thanks!

quentinblampey (approved):
Thanks :)

quentinblampey:
Closes #424

PR #433, opened 2026-07-08
Fix pandas / AnnData 0.13.0 compatibility

alihamraoui:
This PR fixes two failures that occur with newer pandas (copy-on-write) and AnnData 0.13.0.

##### Two changes
- In AnnData 0.13, string indices follow pandas new string dtype, so var_names.values now returns pandas.arrays.ArrowStringArray (1-D, rejects 2-D indexing). Test fixed.

- Read-only array error when zeroing the diagonal → fixed by writing to a writable copy.


quentinblampey (approved):
Perfect, thanks for the quick fix @alihamraoui!

PR #445, opened 2026-08-27
Add CLI command to cache the cellpose model

### LINKED ISSUE #444: [Feature] CLI command to cache the cellpose and/or stardist model
It would be convenient to have a command such as `sopa cache cellpose --pretrained-model ...` (or something like that) to pre-load a model before running cellpose. This can be useful to download the model before running patch segmentation in nf-core/sopa to avoid downloading the model in every single machine.

NB: do we want to have an argument to specify where we store this model? It can be useful for nextflow. Or maybe we just send as an output the location of the cache? We'll need to use this path to pass it to the next process.

alihamraoui:
New `sopa cache cellpose` command to pre-download the Cellpose weights before segmentation (useful for parallel patch segmentation or offline HPC nodes).

It prints the model path (by default `~/.cellpose/models`) to stdout (so it can be used by the workflow manager in `nf-core/sopa` ), and `--model-dir` sets `CELLPOSE_LOCAL_MODELS_PATH`.

I factored the model initialisation out of `cellpose_patch` into `load_cellpose_model`, so the cache and the segmentation share the same logic.

Fixes #444


sopa/cli/cache.py
```diff
@@ -0,0 +1,59 @@
...
+    model_type: str = typer.Option("cyto3", help="Name of the cellpose model if using `cellpose<4.0.0`."),
+    pretrained_model: str = typer.Option(
+        None,
+        help="Name of pretrained model (e.g., `cpsam`, if using `cellpose>=4.0.0`), or path to the pretrained model to be loaded, or `None`.",
+    ),
+    model_dir: str = typer.Option(
+        None,
+        help="Directory where the model is stored. By default, uses the cellpose default directory (`~/.cellpose/models`, or `$CELLPOSE_LOCAL_MODELS_PATH`)",
+    ),
+    gpu: bool = typer.Option(False, help="Whether to instantiate the model on GPU"),
```
quentinblampey: Do we need the gpu argument to download the model? What happens if we always download it on CPU and then load it with a GPU for segmentation?

sopa/cli/cache.py
```diff
@@ -0,0 +1,59 @@
...
+        pretrained_model=pretrained_model,
+        gpu=gpu,
+        cellpose_model_kwargs=method_kwargs,
+    )
+
+    model_path = _cellpose_model_path(model)
+
+    log.info(f"Cellpose model '{pretrained_model or model_type}' is cached and ready to be used.")
+
+    typer.echo(model_path)  # stdout, so it can be captured by a workflow manager for nf-core/sopa
```
quentinblampey: Actually, I'm wondering: since we have the `model_dir` option, maybe we don't need to echo the model-path? Because we already know where it's loaded?

sopa/cli/cache.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def _cellpose_model_path(model) -> str:
+    model = getattr(model, "cp", model)
+
+    return model.pretrained_model
```
quentinblampey: Does it work for both cellpose v3 and v4?

quentinblampey (changes requested):
Thanks @alihamraoui, my comments are mostly questions, but overall the logic is great all, we can do the same for stardist!

sopa/cli/cache.py
alihamraoui: `models.CellposeModel ` takes a gpu argument to save the model on the GPU. I assume this maximizes execution speed. I think it would be useful to expose this in the CLI. do you agree?

sopa/cli/cache.py
alihamraoui: yes, you are right!

sopa/cli/cache.py
alihamraoui: Great question. I just saw that in Cellpose v3, `model.pretrained_model  `returns a `list` of paths. I will adapt it
Thanks

sopa/segmentation/methods/_cellpose.py
```diff
@@ -173,3 +157,48 @@ def _(
...
+        if not pretrained_model:
+            log.info(f"You use cellpose={version}, which requires a `pretrained_model`. Defaulting to 'cpsam'.")
+            pretrained_model = "cpsam"
+
+        if gpu is False:
+            log.warning(
+                f"You use cellpose={version}, which can be slow without a GPU. Consider using `gpu=True`, or downgrading to `cellpose<4.0.0`."
+            )
+
+    return is_v4, pretrained_model
```
quentinblampey, line 180: Do you really need to return `is_v4`? I think it's not used anywhere outside this function, no?

sopa/segmentation/methods/_cellpose.py
```diff
@@ -173,3 +157,48 @@ def _(
         cellpose_model_kwargs=cellpose_model_kwargs,
         **cellpose_eval_kwargs,
     )
+
+
+def _cellpose_version_check(pretrained_model: str | None, gpu: bool | None = None) -> tuple[bool, str | None]:
```
quentinblampey, line 162: Following on my previous comment: if you decide not to return `is_v4`, then the function should probably be named differently (maybe `_cellpose_model_sanity_check` - but it's quite generic, don't hesitate to propose something better)

quentinblampey (changes requested):
Hi @alihamraoui, some minor comments, but everything else looks really great!

sopa/cli/download.py
```diff
@@ -0,0 +1,75 @@
...
+        model_dir_path = Path(model_dir).resolve()
+        model_dir_path.mkdir(parents=True, exist_ok=True)
+        os.environ["KERAS_HOME"] = str(model_dir_path)
+
+    from sopa.segmentation.methods._stardist import SuppressPrintsAndWarnings, load_stardist_model
+
+    log = logging.getLogger(__name__)
+
+    with SuppressPrintsAndWarnings():
+        load_stardist_model(model_type=model_type)
```
quentinblampey: An additional comment: maybe this `SuppressPrintsAndWarnings()` function can be called directly within `load_stardist_model` so that we don't need to import it here?

sopa/segmentation/methods/_cellpose.py
alihamraoui: We still use it in channel check logic, line 144 and 148.

sopa/cli/download.py
alihamraoui: yes 👌🏼 good idea!

sopa/segmentation/methods/_cellpose.py
quentinblampey: Right, sorry I made a `cmd+F` to serach for it on the github UI but these lines were toggled

quentinblampey (approved):
Awesome, thanks a lot :)

Issue #361, opened 2025-12-02
[Feature] Add numpy array mask from external tool to sopa's spatialdata


PietroD:
Dear Quentin,

How would one add the results of an external segmentation tool, whose output is a NumPy array mask, to an existing sdata object? 
Ideally, after adding the mask, I’d like to run the aggregation step and other utilities.

Would simply adding a “custom_boundaries” GeoDataFrame to the sdata.shapes slot work for this purpose?

Thanks

PR #448, opened 2026-09-13
Document how to import an existing segmentation mask

psgundla:
# Document how to import an existing segmentation mask

Related to #361.

The custom segmentation tutorial shows how to run a segmentation function through Sopa, but it does not cover importing a mask produced by another tool. This adds a small example following the approach described in the issue: vectorize the mask, attach its coordinate transformations, and aggregate transcripts using the resulting boundaries.

The example uses a synthetic mask on a toy image and does not require a segmentation model or data download. It uses separate output keys and explains the parts that need care with real inputs: matching pixel size and origin, handling crops or registration, and retaining a verified mapping to source cell IDs. It also notes that polygon conversion is not lossless and that Sopa can auto-save changes to disk-backed objects.

Only the tutorial and its regression test are changed. No production code or dependencies are modified.

### Checks

The new test executes the example cells directly from the notebook. It checks that the source mask, transcripts, and existing shapes remain unchanged, then compares every output gene count with an independent point-in-polygon calculation in the same coordinate frame.

With the development environment installed:

```sh
uv run pytest tests/test_external_mask_docs.py tests/test_vectorize.py -q
```

- Focused tests: **6 passed**.
- Short suite excluding Cellpose, long, and WSI tests: **59 passed, 7 deselected**, with 32 warnings. This run used a temporary cache because the local sandbox blocked Sopa's default home-cache directory; the override was process-local, not a code change.
- Normal docs build, repository-wide pre-commit checks, and `git diff --check`: passed.
- Strict docs build still fails on the existing `docs/README.md` / `docs/index.md` collision. Those files are unchanged.

Tested with Python 3.11.16 and SpatialData 0.7.3.

### Using an existing mask

If you have already segmented an image in another tool, you should not need to run segmentation again to count transcripts in Sopa. This example shows where to bring that mask into the workflow and how to use it for aggregation.

The main thing to check with your own data is alignment: the mask needs the right pixel size, origin, and transformation for the image it came from. It is also worth checking the converted boundaries on the image before counting transcripts. The tutorial points out that output indices are not the original cell IDs, so they should not be used directly to join annotations. The tests use synthetic inputs; this has not been tested on a real dataset.


quentinblampey:
Alright, I added some information about it, please let me know whether you like this version!

psgundla:
Thanks @quentinblampey this version addresses both points clearly. Looks good to me, please go ahead and merge!

quentinblampey (approved):
Sounds good, thanks a lot!

